PR #2, opened 2017-04-10
Form Builder UI

hcarneiro:
@squallstar I'll need your review on the code
@tonytlwu I'll need your review on the UI/UX
Thank you both!

--- 

## UPDATE
**1- Changes**
- [x] Add categories to fields to create an accordion panel for each category and append the fields inside the respective accordion panel
  - [x] Category changes and logic - **NV**
  - [x] Accordion UI - **HC**
- [x] Get name of field to add in the title of the settings overlay - **NV**
  - [x] Replace "Edit field settings" with field name - **HC**
- [x] Change "Required" setting from a select box to radio buttons - **HC**
- [x] Tweak accordions to be contained inside the area - **HC**
- [x] Turn Submit and Clear buttons into fields - **HC**

**(Will be hidden for now)** File field needs to have an option to allow users to select a folder where to save the files that are uploaded - **NV**

**2- Bugs**
- [x] Fields on the right side start to disappear when you **drop them on the right side** - **NV**
http://recordit.co/6PDDh1Xfnm
- [x] Fields are mixed. All fields from different accordion panels appear as different fields in the preview - **NV**
http://recordit.co/GGSlVcVHwF
- [x] Fields from templates don’t have ID - **NV**
http://recordit.co/zHOMB8n5Wr
- [x] Placeholder (a simple form that appears on the page when the component isn't configured and is dropped into the page) doesn't appear after removing the widget from the page and adding a new one from the same component - **NV/HC**
http://recordit.co/knqJlIhux8
- [x] Checkboxes seem to reset Select and Radio - **NV**
http://recordit.co/EiIsDOuEpo\
- [x] Menu data sources appear in the select box in the form settings; They shouldn't - **NV**
http://recordit.co/PkML2bw9NZ
- [x] On creating new data source, the new data source doesn't get appended to the Select box - **HC**
- [x] On Checkbox and Radio settings, in the textarea to write the options, a user can't give a space at the end of a word because of this: https://github.com/Fliplet/fliplet-widget-form-builder/blob/ui-form/js/libs/core.js#L151 - I suggest we only trim at saving, but I'll let you decide - **NV**


hcarneiro:
@squallstar @tonytlwu updated the PR with some tasks needed to be done and bugs needed to resolve.

PR #10, opened 2017-05-23
Online and offline settings

hcarneiro:


js/libs/form.js
```diff
@@ -109,11 +110,25 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
     mounted: function() {
+      var $vm = this;
+
       $(selector).removeClass('is-loading');
 
       Fliplet.Hooks.on('beforeFormSubmit', function(data) {
         console.log('[Hook] beforeFormSubmit', data);
       });
+
+      Fliplet.Navigator.onOnline(function() {
```
squallstar: this can actually be refactored to

```js
if (data.offline === false) {
  Fliplet.Navigator.onOnline(function() {
    $vm.isOffline = false;
  });

  Fliplet.Navigator.onOffline(function() {
    $vm.isOffline = true;
  });
}
```

since there's no point in registering the handlers when the form can be sent in both modes.

hcarneiro: True... Will refactor

js/libs/form.js
```diff
@@ -118,17 +118,15 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
+        });
 
-      Fliplet.Navigator.onOffline(function() {
-        if (!data.offline) {
+        Fliplet.Navigator.onOffline(function() {
           $vm.isOffline = true;
-        }
-      });
+        });
+      }
```
hcarneiro: @squallstar review please!

PR #16, opened 2017-06-01
Feat/email templates

hcarneiro:
Form builder email templates and UI tweaks


js/libs/builder.js
```diff
@@ -191,21 +196,127 @@ var app = new Vue({
...
     'permissionToChange': function(newVal) {
       Fliplet.Widget.toggleSaveButton(newVal);
     },
     'isAddingFields': function(newVal) {
       if (newVal) {
         Fliplet.Studio.emit('widget-mode', 'wide');
-        setTimeout(() => {
-          attatchObservers()
-        }, 1)
+        setTimeout(function() {
```
tonytlwu: Why the change from `() =>{}` to `function() {}`?

hcarneiro: must have copied from somewhere else!

js/libs/builder.js
```diff
@@ -1,7 +1,7 @@
 var data = Fliplet.Widget.getData() || {};
 
 function changeSelectText() {
-  setTimeout(function() {
+  setTimeout(() => {
```
squallstar: don't use ES6, not all devices support it.

js/libs/builder.js
```diff
@@ -83,7 +84,10 @@ var app = new Vue({
       toChangeTemplate: false,
       permissionToChange: false,
       newTemplate: '',
-      redirect: formSettings.redirect
+      redirect: formSettings.redirect,
+      toggleTemplatedEmail: formSettings.onSubmit.indexOf("templatedEmail") > -1 ? true : false,
```
squallstar: single quotes everywhere please :)
also: no need to do `? true : false`, just write the operation (`foo > -1`)

js/libs/builder.js
```diff
@@ -170,8 +175,8 @@ var app = new Vue({
       }
 
       setTimeout(() => {
```
squallstar: no ES6

js/libs/builder.js
```diff
@@ -191,11 +196,117 @@ var app = new Vue({
...
     save: function() {
       return Fliplet.Widget.save(this.settings);
+    },
+    configureEmailTemplate: function() {
+      var $vm = this;
+      var allFields = $vm.settings.fields;
+      // Creates default email template
+      var defaultEmailTemplate = '<h1>' + $vm.settings.displayName + '</h1><p>A new form submission has been received.</p>';
+      defaultEmailTemplate += '<ul>';
+      for (var i = 0; i < allFields.length; i++) {
```
squallstar: can't you just use `$vm.settings.fields.forEach` and avoid using a `for` cycle and `allFields[i]` multiple times?

js/libs/builder.js
```diff
@@ -191,11 +196,117 @@ var app = new Vue({
...
+                hooks: dataSource.hooks
+              });
+            }
+          });
+        }
+
+        Fliplet.Widget.autosize();
+      });
+    },
+    configureGenerateEmailDevice: function() {
```
squallstar: configure or generate?

js/libs/builder.js
```diff
@@ -191,11 +196,117 @@ var app = new Vue({
...
+      var $vm = this;
+      var allFields = $vm.settings.fields;
+      // Creates default email template
+      var defaultEmailTemplate = '<h1>' + $vm.settings.displayName + '</h1><p>A form submission has been received.</p>';
+      defaultEmailTemplate += '<ul>'
+      for (var i = 0; i < allFields.length; i++) {
```
squallstar: pretty sure I've seen this already above. Why is this whole code duplicated?

js/libs/builder.js
```diff
@@ -204,8 +315,8 @@ var app = new Vue({
       if (newVal) {
         Fliplet.Studio.emit('widget-mode', 'wide');
         setTimeout(() => {
-          attatchObservers()
-        }, 1)
+          attatchObservers();
```
squallstar: typo in attach

js/libs/builder.js
```diff
@@ -204,8 +315,8 @@ var app = new Vue({
       if (newVal) {
         Fliplet.Studio.emit('widget-mode', 'wide');
         setTimeout(() => {
```
squallstar: no ES6

js/libs/builder.js
```diff
@@ -234,22 +345,23 @@ var app = new Vue({
       this.settings = generateFormDefaults(settings);
       this.fields = this.settings.fields;
 
-      this.save().then(function() {
+      this.save().then(() => {
```
squallstar: no ES6

js/libs/builder.js
```diff
@@ -234,22 +345,23 @@ var app = new Vue({
...
       });
     },
     'section': function(value) {
+      var $vm = this;
       if (value === 'settings') {
-        changeSelectText()
+        changeSelectText();
 
         if (!this.resultEditor) {
           setTimeout(() => {
```
squallstar: no ES6

js/libs/builder.js
```diff
@@ -258,11 +370,51 @@ var app = new Vue({
...
-          setTimeout(function() {
+          setTimeout(() => {
+            $vm.resultEditor.refresh();
+          }, 1);
+        }
+      }
+    },
+    'settings.onSubmit': function(array) {
+      var $vm = this;
+      this.showDataSource = array.indexOf("dataSource") > -1 ? true : false;
```
squallstar: no need for `?true:false`

js/libs/builder.js
```diff
@@ -353,17 +515,21 @@ var app = new Vue({
...
+            $vm.setupCodeEditor();
+          }, 1);
         }
 
         return;
       }
 
       // Save and close
-      $vm.save().then(function() {
+      $vm.save().then(() => {
```
squallstar: no ES6

js/libs/builder.js
```diff
@@ -353,17 +515,21 @@ var app = new Vue({
...
         Fliplet.Widget.complete();
       });
     }
+
+    Fliplet.API.request('v1/user').then(function(response) {
+      window.userData = response.user;
```
squallstar: why set it to the window? can you keep it private?

hcarneiro: @squallstar I need to use that data later in the code, how can I set the var on a global level to the entire `builder.js` code but keep it private?

js/libs/builder.js
squallstar: @tonytlwu because we don't compile these files with babel, hence using ES6 might fail where it's not supported to I asked to revert to "ol' plain ES5".

PR #19, opened 2017-06-08
Chore/fixes tweaks

squallstar:
- fixes file upload


js/libs/builder.js
```diff
@@ -512,7 +518,7 @@ var app = new Vue({
     }
 
     Fliplet.API.request('v1/user').then(function(response) {
-      userData = response.user;
+      $vm.userData = userData;
```
tonytlwu: Why are we requesting from `v1/user` if the response isn't used at all?

hcarneiro: My bad! mistake.

PR #27, opened 2017-06-27
Show usage on generate email

hcarneiro:


js/libs/builder.js
```diff
@@ -306,6 +306,14 @@ var app = new Vue({
       var $vm = this;
       var emailProviderData = ($vm.settings && $vm.settings.generateEmailTemplate) || $vm.defaultEmailSettingsForCompose;
 
+      emailProviderData.options = {
```
squallstar: is `emailProviderData` always an object? just wondering if can be undefined.

js/libs/builder.js
```diff
@@ -306,6 +306,14 @@ var app = new Vue({
...
+        usage: {
+          'field-x': 'Insert the value entered in the form field.<br><i>To see the ID of each form field, click to edit the field and the ID can be seen at the top right corner.</i>',
+          appName: 'Insert your app name',
+          organisationName: 'insert your organisation name'
```
squallstar: capital letter on `Insert`?

js/libs/builder.js
hcarneiro: Yes it's always a object!

js/libs/builder.js
hcarneiro: Fixed

PR #29, opened 2017-07-28
Editable keys

squallstar:


templates/configurations/form.interface.hbs
```diff
@@ -1,16 +1,23 @@
 <form v-on:submit.prevent="onSubmit">
   <div v-if="_isFormField">
-    <div class="form-group field-id">
+    <!--<div class="form-group field-id">
       <p id="field-id"><span class="field-label">ID:</span> \{{ name }}</p>
+    </div>-->
```
tonytlwu: Maybe we can take this away

squallstar: done

PR #30, opened 2017-07-28
Update dataSource columns on save

squallstar:


hcarneiro:
@squallstar Asked CS and Marketing for a new copy for the tooltip text.
Will update it once I get it, then I can approve!

PR #35, opened 2017-08-16
support for setting the hooks for hashing the password

squallstar:
- Adds a radio box to the password field as per image below
- Removes the default value input box for the password
- Sets hooks based on whether the password needs to be hashed
- Support for password confirmation field 

p.s. I haven't investigated regarding that weird alignment it has got

Img

Hook:

```
{
  type: 'operations',
  runOn: ['beforeQuery', 'beforeSave'],
  payload: {
    password: ['hash']
  }
}
```


squallstar:
merging this so that ricky and I can set up the client app.

PR #41, opened 2017-08-24
Updating settings copy

wispinn:
Some copy in the settings section was confusing to several clients recently, updated to be more specific.


wispinn:
@tonytlwu

PR #42, opened 2017-08-25
field type as static var

squallstar:
ref https://github.com/Fliplet/fliplet-studio/issues/1322

vuejs ref https://github.com/vuejs/vue/issues/6313


squallstar:
@tonytlwu awaiting your reviewal

PR #47, opened 2017-10-04
Fixes legacy issues

hcarneiro:
- Cleanup for legacy options
- Fixes templates to be sent to users

### DIFF
--- a/js/libs/builder.js
+++ b/js/libs/builder.js
@@ -5,9 +5,9 @@ var data = Fliplet.Widget.getData(widgetId) || {};
 if (data.fields) {
   data.fields = _.compact(data.fields);
 }
-if (data.onSubmit && data.onSubmit.length) {
+if (Array.isArray(data.onSubmit) && data.onSubmit.length) {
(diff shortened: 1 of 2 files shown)


js/libs/builder.js
```diff
@@ -5,6 +5,15 @@ var data = Fliplet.Widget.getData(widgetId) || {};
 if (data.fields) {
   data.fields = _.compact(data.fields);
 }
+if (data.onSubmit && data.onSubmit.length) {
```
squallstar: `if (Array.isArray(data.onSubmit) && data.onSubmit.length)`

js/libs/builder.js
```diff
@@ -5,6 +5,15 @@ var data = Fliplet.Widget.getData(widgetId) || {};
...
+  data.onSubmit.forEach(function(el, i) {
+    if(data.onSubmit[i] === 'templatedEmail'){
```
squallstar: `el === 'templatedEmail'` ?

PR #49, opened 2017-10-06
Fix: Email templates and hooks aren't saving

tonytlwu:
Ref: https://github.com/Fliplet/fliplet-studio/issues/1510

### DIFF
--- a/js/libs/builder.js
+++ b/js/libs/builder.js
@@ -96,6 +96,7 @@ var app = new Vue({
       dataSources: [],
       section: 'form', // form or settings
       settings: formSettings,
+      widgetInstanceId: formSettings.id,
       templates: [],
       chooseTemplate: !formSettings.templateId,
       toChangeTemplate: false,
@@ -292,10 +293,10 @@ var app = new Vue({
 
         if ($vm.settings.dataStore.indexOf('dataSource') > -1 || $vm.settings.dataSourceId) {
           var newHook = {
-            widgetInstanceId: $vm.settings.id,
+            widgetInstanceId: $vm.widgetInstanceId,
             runOn: ['insert'],
             type: 'email',
-            payload: $vm.settings.emailTemplateAdd
+            payload: $vm.emailTemplateAdd
           };
 
           Fliplet.DataSources.getById($vm.settings.dataSourceId).then(function(dataSource) {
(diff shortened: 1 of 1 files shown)


js/libs/builder.js
```diff
@@ -299,29 +299,37 @@ var app = new Vue({
...
-            } else {
-              // Add new hook
+            }
+
+            // Update existing hook
+            var currentHook = _.find(dataSource.hooks, function(o) {
+              return o.widgetInstanceId == $vm.settings.id;
+            });
+            
+            if (!currentHook) {
```
tonytlwu: I've assessed the issue as much as I can and it appears that `$vm.settings.id` is `undefined` hence `currentHook` is not found.

This is a problem because even if new hooks are added they wouldn't have a `widgetInstanceId` reference.

Question is: why would `$vm.settings.id` be `undefined`?

tonytlwu: @squallstar Any idea?

squallstar: uhm, I think @hcarneiro changed this code since I first implemented it. I can't remember of `$vm.settings.id`. @hcarneiro does it ring any bell?

@tonytlwu I'm happy to take a look this but I don't think I will be able to before Monday.

hcarneiro:
So I have done as far as I could. The `$vm.settings.id` for some reason disappears after we close the provider, so I created another var that I can use and now we can always access the `widget InstanceId`.

@seromenho @squallstar 
However when creating a new hook like this:
```
var newHook = {
    widgetInstanceId: $vm.widgetInstanceId,
    runOn: ['update'],
    type: 'email',
    payload: $vm.emailTemplateEdit
};

dataSource.hooks.push(newHook);
return Fliplet.DataSources.update($vm.settings.dataSourceId, {
    hooks: dataSource.hooks
});
```
Is creating this: 
[
  {
     "type": "operations",
     "runOn": ["beforeSave", "beforeQuery"],
     "payload": {},
     "widgetInstanceId": 115
  }
]
```
So, even when updating the hook like this:
```
var index = _.findIndex(dataSource.hooks, function(o) {
    return o.widgetInstanceId == $vm.widgetInstanceId;
});
dataSource.hooks.splice(index, 1, currentHook);

return Fliplet.DataSources.update($vm.settings.dataSourceId, {
    hooks: dataSource.hooks
});
```
The hooks haven't changed.

cc @tonytlwu

squallstar:
superseded by https://github.com/Fliplet/fliplet-widget-form-builder/pull/52

PR #53, opened 2017-10-06
wait for hooks update before save

squallstar:
follows https://github.com/Fliplet/fliplet-widget-form-builder/pull/52

ref https://github.com/Fliplet/fliplet-studio/issues/1510

### DIFF
--- a/js/libs/builder.js
+++ b/js/libs/builder.js
@@ -354,7 +354,7 @@ var app = new Vue({
             widgetInstanceId: widgetId,
             runOn: ['update'],
             type: 'email',
-            payload: $vm.emailTemplateEdit
+            payload: JSON.parse(JSON.stringify($vm.emailTemplateEdit))
           };
 
           operation = Fliplet.DataSources.getById($vm.settings.dataSourceId).then(function(dataSource) {
--- a/js/libs/builder.js
+++ b/js/libs/builder.js
@@ -290,6 +290,8 @@ var app = new Vue({
         window.emailTemplateAddProvider = null;
         $vm.emailTemplateAdd = result.data;
 
+        var operation = Promise.resolve();
+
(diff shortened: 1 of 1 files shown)


js/libs/builder.js
```diff
@@ -292,49 +292,35 @@ var app = new Vue({
...
             runOn: ['insert'],
             type: 'email',
-            payload: $vm.emailTemplateAdd
+            payload: JSON.parse(JSON.stringify($vm.emailTemplateAdd))
           };
 
           operation = Fliplet.DataSources.getById($vm.settings.dataSourceId).then(function(dataSource) {
-            var currentHook;
+            // Remove old hook
+            dataSource.hooks = _.reject(dataSource.hooks, { type: 'email', widgetInstanceId: widgetId });
```
hcarneiro: Line 305: This can't be like this, because there might be a hook for the Edit template, and will match the same `type` and `widgetId` and you don't want to delete it.
Use the `runOn` key instead.

js/libs/builder.js
```diff
@@ -363,38 +349,26 @@ var app = new Vue({
...
           };
 
           operation = Fliplet.DataSources.getById($vm.settings.dataSourceId).then(function(dataSource) {
-            if (dataSource.hooks.length) {
-              // Update existing hook
-              var currentHook = _.find(dataSource.hooks, function(o) {
-                return o.widgetInstanceId == $vm.settings.id;
-              });
+            // Remove old hook
+            dataSource.hooks = _.reject(dataSource.hooks, { type: 'email', widgetInstanceId: widgetId });
```
hcarneiro: Line 362: This can't be like like this, because there might be a hook for the Add template, and will match the same `type` and `widgetId` and you don't want to delete it.
Use the `runOn` key instead.

js/libs/builder.js
squallstar: ah true, thanks

PR #61, opened 2017-10-20
Cleaned up form builder UI to be more compact

tonytlwu:
Tightened up padding and layout to be more compact.

### Before

screen shot 2017-10-20 at 20 03 48

### After

screen shot 2017-10-20 at 20 04 00


tonytlwu:
This should give us this for the smaller screens:

screen shot 2017-10-20 at 20 05 21

PR #62, opened 2017-10-24
Fixes for issue #1580

hcarneiro:


js/libs/form.js
```diff
@@ -50,6 +50,10 @@ Fliplet.Widget.instance('form-builder', function(data) {
             } else {
               field.value = moment(entry.data[field.name]).format('DD/MM/YYYY');
             }
+          } else if (field._type === "flCheckbox" && !entry.data[field.name]) {
```
squallstar: the last condition should be replaced with `!Array.isArray(entry.data[field.name])`

PR #66, opened 2017-11-02
only send email when service not bound to a data source

squallstar:
ref https://github.com/Fliplet/fliplet-studio/issues/1556


hcarneiro:
@squallstar what about the following line below those ones?
https://github.com/Fliplet/fliplet-widget-form-builder/blob/master/js/libs/form.js#L238-L244

PR #73, opened 2017-11-17
Forces Vue render to update onInput

tonytlwu:


js/libs/form.js
```diff
@@ -134,6 +134,8 @@ Fliplet.Widget.instance('form-builder', function(data) {
           }
         });
 
+        this.$forceUpdate();
```
squallstar: I'm a bit scared by this. It means the form will be refreshed every single time one of its input changes?

PR #76, opened 2017-11-21
reset events

squallstar:
```js
  form.on('reset', function () {
    console.log('form', form)
    console.log('got value', form.field('Question 1').val())
    console.log('set value', form.field('Question 1').val('2'))
  });
});
```

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -121,6 +121,7 @@ Fliplet.Widget.instance('form-builder', function(data) {
         });
 
         Fliplet.FormBuilder.emit('reset');
+        this.$emit('reset');
       },
       onError: function (fieldName, error) {
         if (!error) {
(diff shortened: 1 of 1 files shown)


js/libs/form.js
```diff
@@ -384,7 +388,11 @@ Fliplet.Widget.instance('form-builder', function(data) {
                 }
 
                 field.value = value;
-                $form.$forceUpdate();
+
+                // Wait for DOM to update, then force a refresh
+                setTimeout(function () {
```
tonytlwu: @squallstar Maybe use https://vuejs.org/v2/api/#Vue-nextTick instead of `setTimeout()`?

squallstar: @tonytlwu I've tried but it doesn't work.

PR #77, opened 2017-11-21
bugfix for progress

squallstar:
ref https://github.com/Fliplet/fliplet-studio/issues/1815


squallstar:
@hcarneiro yeah, looks like I broke it a while ago with a typo.

PR #80, opened 2017-12-01
Style tweaks to select box

hcarneiro:

PR #86, opened 2018-01-02
restructure on organisation templates select

squallstar:
will probably conflict with https://github.com/Fliplet/fliplet-widget-form-builder/pull/80 unless we merge this first and update #80 later.

---

img
  
img2


css/builder.css
```diff
@@ -21,6 +21,24 @@ html {
...
+
+.btn-secondary:hover,
+.btn-secondary:active,
+.btn-secondary:focus,
+.btn-secondary:active:hover,
+.btn-secondary:active:focus {
+  color: #00abd1;
+  background-color: transparent;
+  border-bottom-color: #007994;
+}
```
tonytlwu: Maybe `.btn-secondary` should be added to `fliplet-studio-ui`?

js/libs/builder.js
```diff
@@ -77,13 +77,19 @@ function generateFormDefaults(data) {
     name: '',
     dataSourceId: '',
     templateId: '',
+    previewingTemplate: '',
```
squallstar: what's this? @hcarneiro

js/libs/builder.js
```diff
@@ -703,6 +757,9 @@ var app = new Vue({
         email: data.email,
         type: 'to'
       });
+    },
+    'settings.template': function() {
+      this.editor.setContent(this.settings.description);
```
squallstar: I guess you only need this when it's turned on, right? the first argument in the function should be the value

js/libs/builder.js
hcarneiro: @squallstar That's a flag for when the user is previewing the template.

css/builder.css
```diff
@@ -21,6 +21,24 @@ html {
   text-align: center;
 }
 
+.btn-secondary {
+  color: #00abd1;
+  background-color: transparent;
+  border-bottom-color: #00abd1;
+  border-color: #00abd1;
```
tonytlwu: Not sure why `border-bottom-color` and `border-color` are needed

css/builder.css
```diff
@@ -21,6 +21,24 @@ html {
...
+}
+
+.btn-secondary:hover,
+.btn-secondary:active,
+.btn-secondary:focus,
+.btn-secondary:active:hover,
+.btn-secondary:active:focus {
+  color: #00abd1;
+  background-color: transparent;
+  border-bottom-color: #007994;
```
tonytlwu: `border-bottom-color` is the only one that's changed from the above. Are the other 2 property declarations needed?

PR #88, opened 2018-01-05
Fixes

hcarneiro:
Fixes:
https://github.com/Fliplet/fliplet-studio/issues/1958
https://github.com/Fliplet/fliplet-studio/issues/1948

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -5,7 +5,7 @@ Fliplet.Widget.instance('form-builder', function(data) {
   var selector = '[data-form-builder-id="' + data.id + '"]';
   var progressKey = 'form-builder-progress-' + (data.uuid || data.id);
 
-  var entryId = Fliplet.Env.get('mode') !== 'interact' && data.dataSourceId && Fliplet.Navigate.query.dataSourceEntryId;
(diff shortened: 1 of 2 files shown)


js/libs/form.js
```diff
@@ -5,7 +5,7 @@ Fliplet.Widget.instance('form-builder', function(data) {
   var selector = '[data-form-builder-id="' + data.id + '"]';
   var progressKey = 'form-builder-progress-' + (data.uuid || data.id);
 
-  var entryId = data.dataSourceId && Fliplet.Navigate.query.dataSourceEntryId;
+  var entryId = Fliplet.Env.get('mode') !== 'interact' && data.dataSourceId && Fliplet.Navigate.query.dataSourceEntryId;
```
seromenho: `Fliplet.Env.get('mode') !== 'interact'` same as `!Fliplet.Env.get('interact')`

hcarneiro: @seromenho Committed the change

seromenho: @hcarneiro I wasn't suggesting the change. Either way was ok. It's the same. It was only because the conversation you asked me and then I saw you made another way, I was now just saying it's the same.

PR #93, opened 2018-01-22
new events and listeners

squallstar:
Two new methods added to form fields:

- `change(fn [,runOnBind = true])` registers a function to be called whenever the field value changes
- `toggle(bool)` shows and hides the field, also reverting its value to the default value. Field will also not be posted and progress won't be saved

```js

  // registers a callback to be fired whenever the field value changes
  form.field('Question 1').change(function (val) {

    // only show / enable this field when val is "foo"
    // works like jQuery "toggle", which evaluates the argument as a "truthy" boolean.
    form.field('Question 2').toggle(val === 'foo');

  });

});
```


squallstar:
@tonytlwu something we could add in the future is a higher level API:

```js
form.field(['Question 2', 'Question 3']).showWhen('Question1', true)
```

PR #110, opened 2018-03-26
time picker

squallstar:
img


hcarneiro:
Updated the time icon @squallstar 
screen shot 2018-03-26 at 16 43 34

PR #117, opened 2018-04-10
Support for natural numbers and iOS numeric keyboard

squallstar:
Displays the numeric keyboard for "number form fields" when only positive numbers are allowed.

---

ref https://github.com/Fliplet/fliplet-studio/issues/2465

---

https://www.filamentgroup.com/lab/type-number.html

img


templates/configurations/number.interface.hbs
```diff
@@ -6,3 +6,22 @@
...
 </div>
+<div class="form-group cleafix">
+    <div class="col-xs-12">
+      <label class="control-label">What numbers should be allowed?</label>
+    </div>
+    <div class="col-xs-12">
+      <div class="radio radio-icon">
+        <input type="radio" id="numbers-all" name="numbers-option" v-bind:value="false" v-model="positiveOnly">
+        <label for="numbers-all">
+          <span class="check"><i class="fa fa-circle"></i></span> <strong>Allow all numbers</strong> (positive and negative integers)
```
tonytlwu: @squallstar Does "(positive and negative integers)" mean floating points are not allowed?

squallstar: @tonytlwu actually all numbers will be allowed because the iOS keyboard allows you to type both `-` and `.`. What do you suggest we write? Simply keep "Allow all numbers"?

tonytlwu: Yeh I think the 2 options written as below should do, without additional explanations:

- Allow all numbers
- Positive numbers only

squallstar:
@tonytlwu done img

PR #122, opened 2018-05-01
fixes for attaching files

squallstar:
ref https://github.com/Fliplet/fliplet-studio/issues/2549

### DIFF
--- a/js/components/file.js
+++ b/js/components/file.js
@@ -29,29 +29,28 @@ Fliplet.FormBuilder.field('file', {
     mediaFolderNavStack: {
       type: Array,
       default: []
+    },
+    value: {
+      type: Array,
+      default: []
     }
   },
   computed: {
     selectedFileName: function() {
-      var names = [];
-
-      for (var i = 0; i < this.selectedFiles.length; i++) {
-        var file = this.selectedFiles.item(i);
-        names.push(file.name);
-      }
-
-      return names.join(', ');
+      return _.map(this.selectedFiles, 'name').join(', ');
     }
   },
   methods: {
     updateValue: function() {
       var $vm = this;
-      this.selectedFiles = this.$refs.fileInput.files;
+
+      this.selectedFiles.splice(0, this.selectedFiles.length);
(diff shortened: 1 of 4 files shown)


js/components/file.js
```diff
@@ -29,29 +29,28 @@ Fliplet.FormBuilder.field('file', {
...
-      return names.join(', ');
+      return _.map(this.selectedFiles, 'name').join(', ');
     }
   },
   methods: {
     updateValue: function() {
       var $vm = this;
-      this.selectedFiles = this.$refs.fileInput.files;
+
+      this.selectedFiles.splice(0, this.selectedFiles.length);
```
tonytlwu: What's the logic and necessity of this particular part? Emptying the array by splicing seems like an odd method 🤔

templates/components/file.build.hbs
```diff
@@ -2,8 +2,8 @@
   <i class="fa fa-plus" aria-hidden="true"></i>
   <span v-if="selectedFileName">Replace file</span>
   <span v-else>Choose file</span>
-  <input type="file" ref="fileInput" :id="name" :name="name" :data-folder-id="mediaFolderId" class="input-file selectfile" :required="required" v-on:change="updateValue()">
+  <input type="file" ref="fileInput" :id="name" :name="name" :data-folder-id="mediaFolderId" class="input-file selectfile" :required="required" v-on:change="updateValue()" multiple>
```
tonytlwu: I also find that when the field is "required", validation triggers an error because the `[type="file"]` field is hidden with `display: none`, which makes the field **unfocusable**. Is it easy to fix this too?

squallstar: @tonytlwu fixed https://github.com/Fliplet/fliplet-widget-form-builder/pull/122/commits/a71bb85c5fc1e854d9450f4dd290c2cf927623ed

js/components/file.js
tonytlwu: I see. Fine with me 🆗

PR #124, opened 2018-05-23
public method to set up form field options

squallstar:
```js
form.field('foo').options([ 'hello', 'world'  ]);

form.field('foo').options([ { id: 'hello', disabled: true }, { value: 'world' }  ]);
```

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -503,8 +503,20 @@ Fliplet.Widget.instance('form-builder', function(data) {
                 }
               },
               options: function (values) {
-                field.options = values.map(function (value) {
-                  return typeof value === 'object' ? value : { id: value };
+                if (!Array.isArray(values)) {
+                  throw new Error('Options must be an array');
+                }
+
+                field.options = values.map(function (option) {
+                  if (typeof option === 'object') {
+                    if (typeof option.value !== 'undefined') {
+                      option.id = option.value;
+                    }
+
+                    return option;
+                  }
+
+                  return { id: option };
                 });
               }
             };
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -501,6 +501,11 @@ Fliplet.Widget.instance('form-builder', function(data) {
                     }
                   });
                 }
+              },
+              options: function (values) {
+                field.options = values.map(function (value) {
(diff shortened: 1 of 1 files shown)


tonytlwu:
@squallstar Can you confirm that this works for all following field types?

- [ ] Checkboxes
- [ ] Radio buttons
- [ ] Dropdowns

squallstar:
@tonytlwu yes although the `disabled` prop is only implemented on dropdowns.

tonytlwu:
@squallstar 

1. Is it much more work to have `disabled` working for the other 2 types too?
1. In this case, I have a slight preference to use `value` instead of `id`

squallstar:
@tonytlwu 

1. I would skip this for now since nobody is asking for it hence it's not justified work. Yes, it will take some time to be tested properly across all use cases and platforms.
2. The form builder internally uses `id` because **Vue** suggested using for speed that instead of other attributes. If needed we can do a remapping before assigning the object but the point is that you should only provide an array of strings (or numbers) unless we provide support for a `label` in the future. We currently don't have this so we shouldn't be using the `id` object syntax anyway.

tonytlwu:
@squallstar 

I realised that `disabled` for checkboxes and radio are way less common anyway, so that's fine.

I suggest doing a mapping between `value` and `id`. When I heard you said:

> The form builder internally uses `id` because **Vue**

...all I could think of was...

> Developers shouldn't need to know how it's built.

Yes I agree the following scenario will be used more:

```js
form.field('foo').options([ 'hello', 'world'  ]);
```

But since you've suggested this syntax too:

```js
form.field('foo').options([ { id: 'hello', disabled: true }, { id: 'world' }  ]);
```

...I suggest using...

```js
form.field('foo').options([ { value: 'hello', disabled: true }, { value: 'world' }  ]);
```

...because it'd make more sense. Other attributes such as `label`, as you said, can be added later.

squallstar:
@tonytlwu done, you can now use either `id` or `value`

tonytlwu:
@squallstar I've changed the description to use only `value`. I understand `id` is allowed but having documentation that says *you can use any of these* only confuses people :)

PR #132, opened 2018-08-07
Feature: new wysiwyg field powered by tinymce

squallstar:
screen shot 2018-08-07 at 19 21 03


tonytlwu:
Cleaned up toolbar

image

...with the following features: Bold, Italic, Underline, Aligh left/center/right, Justify, Bullet list, Numbered list, Outdent, Indent, LTR, RTL, Insert link, Remove format, Code editor, Fullscreen.

I've tried using Mobile mode, but it doesn't seem to work very well. I'll try upgrading to 4.8.1 to see if the latest version is better.

ibroom:
Tony, I think air-mode is designed for mobile. Can you check it? https://summernote.org/examples/#air-mode

Alternatively we only offer this on mobile:
Bold, Italic, Underline, Bullet list, Numbered list, Outdent, Indent, Remove format

Regarding LTR/RTL - does this default to the current app's default from CSS or would an arabic user have to set it each time manually?

tonytlwu:
After upgrading to TinyMCE 4.8.1 Mobile Mode works better now and is by default using a shorter list of toolbar buttons.

_Default_

simulator screen shot - iphone 7 - 2018-08-08 at 11 17 32

_Editor activated_

simulator screen shot - iphone 7 - 2018-08-08 at 11 17 37

PR #162, opened 2019-03-26
support for automatically generate a password

squallstar:
### Adds a new option for the password field to automatically generate a password

When this new option is enabled:
- the field will be set to readonly
- the value of the field will be autogenerated to a random password with the length defined by the interface settings
- password gets only generated when field value is empty
- the password cannot be seen or copied
- when the option is enabled you cannot enable the option for displaying the password confirmation field

**This option is made to be used by Admin/CMS apps to create new users with a random password** and most likely in conjunction with the hashing option turned ON and an email template sent when new entries are added.

Note: the plaintext password can be retrieved in the `beforeFormSubmit` hook. Right after that, the password gets hashed by the backend once received and cannot be read as plaintext anymore aside from the email hook (if set).

img


tonytlwu:
@squallstar Can you expand on how this feature works and what this would mean downstream for the user and system?

templates/components/password.build.hbs
```diff
@@ -1,6 +1,9 @@
 <input
   type="password"
   class="form-control"
+  :class="{ disabled: autogenerate }"
```
tonytlwu: This is a minor point and I don't mind having the `disabled` class here, but I thought having `readonly` means it won't be interactable anyway. Is there any reason this class is necessary?

templates/components/password.build.hbs
```diff
@@ -1,6 +1,9 @@
...
+  :readonly="autogenerate"
+  autocomplete="off"
   v-model.trim="value"
```
tonytlwu: I assume spaces are allowed in passwords, so we shouldn't be trimming the value?

templates/configurations/password.interface.hbs
```diff
@@ -20,6 +20,27 @@
 </div>
 
 <div class="form-group">
+  <label>Should the system autogenerate the password?</label>
```
tonytlwu:
```suggestion
  <label>Should a password be automatically generated when submitting a new entry?</label>
```

templates/configurations/password.interface.hbs
```diff
@@ -20,6 +20,27 @@
...
+  <div class="radio radio-icon">
+    <input type="radio" id="autogenerate-yes" name="autogenerate-option" v-bind:value="true" v-model="autogenerate">
+    <label for="autogenerate-yes">
+      <span class="check"><i class="fa fa-circle"></i></span> <strong>Yes</strong> - Fliplet will autogenerate a secure random password
```
tonytlwu:
```suggestion
      <span class="check"><i class="fa fa-circle"></i></span> <strong>Yes</strong> - a secure password will be randomly generated
```

templates/configurations/password.interface.hbs
```diff
@@ -20,6 +20,27 @@
...
+    </label>
+  </div>
+  <div class="radio radio-icon">
+    <input type="radio" id="autogenerate-no" name="autogenerate-option" v-bind:value="false" v-model="autogenerate">
+    <label for="autogenerate-no">
+      <span class="check"><i class="fa fa-circle"></i></span> <strong>No</strong> - The user will manually input a password
```
tonytlwu:
```suggestion
      <span class="check"><i class="fa fa-circle"></i></span> <strong>No</strong> - The user will manually enter a password
```

templates/configurations/password.interface.hbs
```diff
@@ -20,6 +20,27 @@
...
+    </label>
+  </div>
+</div>
+
+<div class="form-group" v-if="autogenerate">
+  <label>Autogenerated password length</label>
+  <input class="form-control" type="number" v-model="autogenerateLength" placeholder="Password length" min="1" max="32" step="1"/>
```
tonytlwu:
```suggestion
  <input class="form-control" type="number" v-model="autogenerateLength" placeholder="Password length" min="1" max="32" step="1" :required="autogenerate" />
```

js/components/password.js
```diff
@@ -11,6 +11,14 @@ Fliplet.FormBuilder.field('password', {
       type: Boolean,
       default: false
     },
+    autogenerate: {
```
tonytlwu: When user enables `autogenerate`, if the interface has `confirm` enabled, it should trigger an alert with `Fliplet.Modal.alert()` that says `Password confirmation and autogeneration can't both be enabled. Enabling one disables the other.` and then disables the `confirm` setting. And vice versa.

I'm following a convention I saw on the phone the other day that handles these "one or the other" settings.

IMG_99C6F4656143-1

templates/components/password.build.hbs
```diff
@@ -9,7 +12,7 @@
   :required="required"
 />
```
tonytlwu: I'm blinding coding HTML here, but I think when a password is autogenerated and used, the build should include a message below the field.
`<p class="help-block">A password will be automatically generated</p>`

templates/components/password.build.hbs
squallstar: @tonytlwu adding the class seemed to make the input background grey, so it's more obvious about it being readonly.

templates/components/password.build.hbs
squallstar: I didn't change this but I'm OK in getting rid of it.

js/components/password.js
squallstar: @tonytlwu when the autogenerate setting is enabled the password confirmation option does not get displayed in the interface, so you can't enable both. 

I've added the checks in the build.js to ensure we deal with such scenario should it be misconfigured somehow (e.g. interface changes in the future).

tonytlwu: I know. But, the interface should reflect this and toggle the settings accordingly.

templates/components/password.build.hbs
tonytlwu: I know. I just spotted it. Let's get rid of the trim please.

templates/components/password.build.hbs
tonytlwu: CSS should be using `[readonly]` to color the background grey. Let's leave the `disabled` class in for this purpose. Will update the styling selectors in the future so we're not relying on the wrong class to accomplish this.

squallstar:
@tonytlwu note added: 
image

js/components/password.js
squallstar: @tonytlwu added 
image

squallstar:
added modal check as requested
image

templates/components/password.build.hbs
```diff
@@ -1,15 +1,19 @@
...
+  autocomplete="off"
   v-on:input="updateValue()"
   :name="name"
   :id="name"
   :placeholder="placeholder"
   :required="required"
 />
 
-<div class="form-group row clearfix" v-if="confirm" :class="{ 'has-error': hasConfirmationError }">
+<p class="help-block" v-if="autogenerate">A password will be automatically generated</p>
```
tonytlwu: I thought the password field would have asterisks or dots in them. But it looks like you're not loading the field with the auto-generated passwords. I think you're doing this for security purposes so that the value can't be read? In this case, perhaps it's the easiest to just use `placeholder="A password will be automatically generated"` when `autogenerate` is enabled.

js/components/password.js
tonytlwu: @squallstar How about the _vice versa_ part?

If `autogenerate` is enabled and user enables `confirm`, then the reverse should also happen, with the same alert message displayed.

templates/components/password.build.hbs
tonytlwu: That'll mean one less piece of UI on the screen, simpler design and less UI for theme customisation.

js/components/password.js
squallstar: @tonytlwu you can't enable `confirm` when `autogenerate` is enabled because that UI is hidden on such case.

tonytlwu: That works for me 👍

templates/components/password.build.hbs
squallstar: @tonytlwu done 
image

templates/components/password.build.hbs
```diff
@@ -1,19 +1,16 @@
 <input
   type="password"
   class="form-control"
-  :class="{ disabled: autogenerate }"
```
tonytlwu: If you're making this change, the following changes should probably be made too:

**Relevant to this PR**

https://github.com/Fliplet/fliplet-widget-form-builder/blob/master/css/form.css#L276 Add `.form-control[readonly]` to selector

**Irrelevant to this PR**

https://github.com/Fliplet/fliplet-widget-form-builder/blob/master/css/form.css#L498-L505 Add `.select-proxy-display>.hidden-select[readonly]~.icon` and `.select-proxy-display>.hidden-select[ready]~.select-value-proxy` to the selectors.

PR #167, opened 2019-04-11
better validation on input number

squallstar:
### When all numbers are allowed:

- the input type will be set to `text`
- the input will only allow numbers, `-` and `.`

### When positive numbers only are allowed

- the input type will be set to `number`
- the input will only allow numbers and `.` (this is for web only)

### DIFF
--- a/js/components/number.js
+++ b/js/components/number.js
@@ -16,13 +16,11 @@ Fliplet.FormBuilder.field('number', {
   },
   methods: {
     updateValue: function () {
-      var $vm = this;
-
-      var rx = new RegExp(
+      var ensureNumberRx = new RegExp(
         this.positiveOnly ? '[^0-9\.,]' : '[^0-9\.,-]'
-      , 'g')
+      , 'g');
 
-      this.value = this.value.replace(rx, '');
+      this.value = this.value.replace(ensureNumberRx, '');
       this.$emit('_input', this.name, this.value);
     }
   },
--- a/js/components/number.js
+++ b/js/components/number.js
@@ -14,6 +14,47 @@ Fliplet.FormBuilder.field('number', {
       default: 0
     }
   },
+  methods: {
+    verifyInput: function(event) {
+      if (!event) {
+        return;
+      }
+
+      var code = event.which ? event.which : event.keyCode;
+
+      // Do not allow dot/comma characters if settings are not allowing decimals
+      if ([188, 190].indexOf(code) !== -1) {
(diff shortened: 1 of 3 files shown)


tonytlwu:
@squallstar Looks good so far. Be sure to remove the `,` and `.` when the form is submitted so that the field value can be saved/submitted as a `number`.

js/components/number.js
```diff
@@ -14,7 +14,20 @@ Fliplet.FormBuilder.field('number', {
...
     }
   },
+  methods: {
+    updateValue: function () {
+      var ensureNumberRx = new RegExp(
+        this.positiveOnly ? '[^0-9\.,]' : '[^0-9\.,-]'
+      , 'g');
+
+      this.value = this.value.replace(ensureNumberRx, '');
+      this.$emit('_input', this.name, this.value);
```
tonytlwu: So, anything that isn't in the pattern `[^0-9\.,]` or `[^0-9\.,-]` are excluded. That's good.

When the data is submitted to the server, will the `,` be removed so that the `,` is only used for display purposes?

Having `,` saved in the data sources could create problems if there aren't currently sufficient type checking/casting in place.

squallstar: @tonytlwu good point, should I just `.replace(/,/g, '.')` here on line 23?

tonytlwu: I imagine it'll be `.replace(/,/g, '')` based on UK/US standards, but the challenge, as your example reminds me, is that `.replace(/.//g, '')` is what should be happening for EU users.

Not sure how to best resolve this.

squallstar: well, this replacement happens on each keypress, meaning that when you type a comma it gets replaced to a dot instantly. That would should be quite obvious what the meaning is? What do you think?

Also: keep in mind we currently don't do any replacement so it's potentially a breaking change for companies like Socotec which are managing this on their own.

tonytlwu: If when a user pastes `12,300,230.00` it becomes `12.300.230.00` I think:

1. It's a little confusing. User can stare at it and work it out after a while, but...
1. User could end up just submitting it as it is as there aren't any more sophisticated validations in place

My lazy brain tells me `.replace(/,/g, '')` should do the trick. `parseFloat('12,300')` currently gives me `12` so we should probably do that anyway.

squallstar: @tonytlwu I've basically updated the initial regexp to not include `,` :-)

squallstar: @tonytlwu and it works great 

gif

PR #170, opened 2019-05-14
ensure on load entry error, message gets displayed

squallstar:
ref https://github.com/Fliplet/fliplet-studio/issues/4328#issuecomment-492150670

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -420,7 +420,9 @@ Fliplet.Widget.instance('form-builder', function(data) {
             $vm.error = error;
             $vm.isLoading = false;
 
-            Fliplet.UI.Toast.error(error);
+            Fliplet.UI.Toast.error(error, {
+              message: 'The entry to edit could not be loaded.'
+            });
           });
         }
 
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -421,7 +421,7 @@ Fliplet.Widget.instance('form-builder', function(data) {
             $vm.isLoading = false;
 
             Fliplet.UI.Toast.error(error, {
-              message: 'The entry to edit could not be loaded.'
+              message: 'Unable to load entry'
             });
           });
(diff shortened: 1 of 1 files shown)


js/libs/form.js
```diff
@@ -416,7 +416,11 @@ Fliplet.Widget.instance('form-builder', function(data) {
             $vm.fields = getFields();
             $vm.isLoading = false;
           }).catch(function (err) {
-            $vm.error = err.message || err.description || err;
+            var error = Fliplet.parseError(err);
+            $vm.error = error;
+            $vm.isLoading = false;
+
+            Fliplet.UI.Toast.error(error);
```
tonytlwu:
```suggestion
            Fliplet.UI.Toast.error(error, {
              message: defaultMessage
            });
```

Without a message, it'll say "Unexpected error" with a "Details" link to learn more. Is that suitable for this scenario or is there a more appropriate message that can be used instead of "Unexpected error"?

squallstar: @tonytlwu thanks! fixed

js/libs/form.js
```diff
@@ -416,7 +416,13 @@ Fliplet.Widget.instance('form-builder', function(data) {
             $vm.fields = getFields();
             $vm.isLoading = false;
           }).catch(function (err) {
-            $vm.error = err.message || err.description || err;
+            var error = Fliplet.parseError(err);
+            $vm.error = error;
+            $vm.isLoading = false;
+
+            Fliplet.UI.Toast.error(error, {
+              message: 'The entry to edit could not be loaded.'
```
tonytlwu: Since it's always about editing entries and users would have likely got here via some UI/interaction that implies "editing", perhaps the "edit" context would be clear to the user and we can simplify the message:

```suggestion
              message: 'Unable to load entry'
```

Thoughts?

PR #177, opened 2019-06-03
upload files as binary through multipart

squallstar:
ref https://github.com/Fliplet/fliplet-studio/issues/3783

---

requires https://github.com/Fliplet/fliplet-api/pull/3378


squallstar:
just put this on staging for testing

PR #181, opened 2019-06-14
Image issue 3785

SachinSharma81094:
## Issue
User can submit images with larger size than the image field max size settings #3785

## Description
Showing validation error when user try to submit images with larger size than the image field max size settings.

## Notes

### DIFF
--- a/js/components/image.js
+++ b/js/components/image.js
@@ -42,7 +42,8 @@ Fliplet.FormBuilder.field('image', {
   data: {
     boundingRect: undefined,
     cameraSource: undefined,
-    forcedClick: false
+    forcedClick: false,
+    isImageValid: undefined
   },
   created: function() {
     Fliplet.FormBuilder.on('reset', this.onReset);
@@ -67,6 +68,10 @@ Fliplet.FormBuilder.field('image', {
       $(this.$refs.imageInput).parents('.form-group').removeClass('has-error');
 
       if (!this.required) {
+        if(this.value.length && !this.isImageValid){
+          $(this.$refs.imageInput).parents('.form-group').addClass('has-error');
+          return Promise.reject('Field [' + this.name + '] height and width exceeded.');
+        }
         return;
       }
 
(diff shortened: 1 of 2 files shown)


js/components/image.js
```diff
@@ -67,6 +68,10 @@ Fliplet.FormBuilder.field('image', {
       $(this.$refs.imageInput).parents('.form-group').removeClass('has-error');
 
       if (!this.required) {
+        if (this.value.length && !this.isImageValid) {
+          $(this.$refs.imageInput).parents('.form-group').addClass('has-error');
```
squallstar: the `$(this.$refs.imageInput).parents('.form-group').addClass('has-error');` line is duplicated, is there a way we can avoid this? e.g. using the `isImageValid` boolean to bind the class via the vue template?

SachinSharma81094: we will look into it.

js/components/image.js
```diff
@@ -75,6 +80,10 @@ Fliplet.FormBuilder.field('image', {
 
         return Promise.reject('Please fill in required fields.');
       }
+      if (!this.isImageValid) {
```
tonytlwu: Empty line before block

```suggestion

      if (!this.isImageValid) {
```

SachinSharma81094: Fixed.

js/components/image.js
```diff
@@ -42,7 +42,8 @@ Fliplet.FormBuilder.field('image', {
   data: {
     boundingRect: undefined,
     cameraSource: undefined,
-    forcedClick: false
+    forcedClick: false,
+    isImageValid: undefined
```
squallstar: this could default to `false`

js/components/image.js
```diff
@@ -146,6 +156,17 @@ Fliplet.FormBuilder.field('image', {
...
+
+      var image = new Image();
+      image.onload = function () {
+        if (this.height > $vm.customHeight || this.width > $vm.customWidth) {
+          $vm.isImageValid = false;
+        } else {
+          $vm.isImageValid = true;
+        }
+      };
+      image.src = window.URL.createObjectURL(file);
```
squallstar: Can you wrap `file` into `Fliplet.Media.authenticate(file)`? I think it might be required to add the authorization token to the files if their URL is remote.

PR #183, opened 2019-06-20
Set values for the options

manst:
@squallstar 
When configuring a select, option or checkbox field add the ability to set a value not just a name

No ref to the GitHub issue

### DIFF
--- a/js/libs/core.js
+++ b/js/libs/core.js
@@ -268,16 +268,32 @@ Fliplet.FormBuilder = (function() {
       // If options is an array, automatically deal with options
       if (hasOptions) {
         component.computed._options = function generateOptions() {
-          return this.options.map(function(option) {
-            return option.id;
+          return this.options.map(function (option) {
+            if (option.id && option.id != option.label) {
+              return option.label + ' <' + option.id + '>';
+            }
+            return option.label;
           }).join('\r\n');
         };
 
         component.methods._setOptions = function setOptions(str) {
-          this.options = _.compact(str.split(/\r?\n/).map(function(s) {
-            return {
-              id: s.trim()
-            };
+          this.options = _.compact(str.split(/\r?\n/).map(function (rawOption) {
+            rawOption = rawOption.trim();
+
+            var regex = /<.*>$/g;
+            var match = rawOption.match(regex);
+            var option = {};
+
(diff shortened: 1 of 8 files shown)


js/libs/core.js
```diff
@@ -268,16 +268,31 @@ Fliplet.FormBuilder = (function() {
...
+            var regex = /<.*>$/g;
+            var match = rawOption.match(regex);
+            var option = {};
+
+            if (match) {
+              option.label = rawOption.replace(regex, '').trim();
+              var value = match[0].substring(1, match[0].length - 1).trim();
+              option.id = value || option.label;
+            } else {
+              option.label = option.id = rawOption;
```
squallstar: split this into two separate assignments please

templates/components/select.build.hbs
```diff
@@ -2,7 +2,7 @@
   <select class="form-control hidden-select" :name="name" :id="name" v-model="value" :required="required" v-on:change="updateValue()">
     <option v-if="placeholder" value="">\{{ placeholder }}</option>
     <option v-for="option in options" :value="option.id" :disabled="option.disabled">
-      \{{ option.id }}
+      \{{ option.label }}
```
squallstar: what about existing forms that in their settings have only the `id`? We need to support those too for existing apps even if they don't go through the process of opening the form interface and saving changes. Does using a fallback work here? e.g. `{{ option.label || option.id }}`

manst: Are there any cases how I can test this behavior with my changes?

js/libs/core.js
manst: Fixed

templates/components/select.build.hbs
squallstar: 1. setup a form locally before making your changes
2. switch the branch to include your changes
3. refresh the output iframe (http://localhost:3000/build) and see if works

You can also test out in a similar way on staging:

1. publish your form component without your changes
2. drop it into a screen and set it up as required
3. publish your form component with the changes (under the same package name)
4. refresh the screen (without making any change via the interface) and verify if it works

manst: Fallback works. Have updated commit

PR #184, opened 2019-06-20
Hide data source settings

manst:
@squallstar 
Form settings related to a data source should only be available when a data source has been selected

No ref to the Github issue


squallstar:
@manst could you please:

1. create a label titled "When the form is loaded..." to be placed before the "Load the data from user's profile" checkbox?  Otherwise it look weird now that by default there's only one checkbox and the header is not relevant to it.

2. Only show the "When the form is submitted..." when the data source is selected.

PR #187, opened 2019-06-24
#3787 When uploading a file then loading the form again the file field remembers the previous file and it shouldn’t 

leraradinovich:
@squallstar

### DIFF
--- a/js/components/file.js
+++ b/js/components/file.js
@@ -40,20 +40,28 @@ Fliplet.FormBuilder.field('file', {
       return _.map(this.selectedFiles, 'name').join(', ');
     }
   },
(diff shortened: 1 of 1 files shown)


squallstar:
Does this break the behaviour when editing a data source entry?

This is how you can test *o*n staging only**:

1. set up the form settings to allow "Update existing entries in a data source"
2. set up a new screen with a "List from Data Source" component, listing the items from the same data source as the one used by the form to insert entry. Also configure the editing capabilities targeting the screen with the form:

image

3. using the form, insert a new entry with images too
4. from the screen with the list, try editing the entry and verify if the image field lists the images previously uploaded

PR #199, opened 2019-07-09
Multiple forms with the same field IDs cannot be on the same screen

Fightnight7:
@squallstar

Added new component.computed._elementId method in core.js. Replaced name with _elementId in the templates of components that have id property.

ref https://github.com/Fliplet/fliplet-studio/issues/4158

Upplabs-28-1
Upplabs-28-2
Upplabs-28-3


js/libs/core.js
```diff
@@ -85,6 +85,10 @@ Fliplet.FormBuilder = (function() {
         return this.showLabel || this.showLabel === undefined;
       };
 
+      component.computed._elementId = function() {
+        return this.name + '-' + this.$parent.$el.dataset.formBuilderId;
```
squallstar: where is `formBuilderId` defined?

squallstar:
@tonytlwu I agree with you, this is indeed a breaking change for users having custom code targeting specific fields by ID, e.g. `$('#email').hide()` and although it's not what we recommend for interacting with the form builder we're almost sure many apps could break after this change.

Furthermore, I'm currently unsure about how the `formBuilderId` in this PR is generated (I've raised a question) but we need to make sure it's consistent between master and production apps.

---

I think these are our options:

1. upgrade the form builder version (e.g. to 1.1) to ensure only new instances will have their IDs changed
2. update the code in this PR to only apply the new format given a specific setting exists in the widget instance; such setting will only be set when creating a form, so we can keep the same version but have different behaviours based on this
3. as you pointed out, investigate if there's any other way we can resolve this without having to change the IDs. This is something @Fightnight7 can look into.

js/libs/core.js
MaksymShUpplabs: It is defined in form.js.
Upplabs #28 answer 1

MaksymShUpplabs:
I am not sure how it is possible to assign unique ID to elements without changing the actual ID.

squallstar:
@MaksymShUpplabs to confirm, this is the issue from the user's perspective 
image

MaksymShUpplabs:
@squallstar could you please clarify what do you mean by "the issue from the user's perspective"?

squallstar:
@MaksymShUpplabs the issue is that when clicking on a label for a input repeated with the same ID on the same page flags checks/unchecks both. Is there any way we can use Javascript to control how the labels are handling the clicks so that we can trigger events only on the inputs for the same form?

MaksymShUpplabs:
In the task description @tonytlwu mentioned that our objective is to have unique IDs for all fields regardless of their types. Do we have now to remove unique IDs generation and just solve the unwanted behaviour with clicking on the radiobuttons and checkboxes input labels?

squallstar:
@MaksymShUpplabs let me have a chat with @tonytlwu and confirm, I'll get back to you.

squallstar:
@MaksymShUpplabs here's an update: we would prefer a JS (Events) based approach instead of actually changing the field IDs because of the latter being a risky breaking change. Can you look into solving this by using JS events?

squallstar:
@MaksymShokin is this PR still relevant?

MaksymShokin:
This PR is not relevant. The issue was solved in another PR.

PR #214, opened 2019-07-23
Added check for corrupted images

MaksymShokin:
@squallstar

Fixed behavior:
The issue was that the user did not see any error when submitting corrupted images.
Now the user will see notification error.

ref https://github.com/Fliplet/fliplet-studio/issues/#4454 

Corrupted image demo


squallstar:
@MaksymShokin ref was incorrect (there was a `#` in the middle of the URL)

PR #220, opened 2019-08-07
Fixes an issue where Rich Text field wasn't submitting when there's a Datepicker in the same form

YaroslavOvdii:
@squallstar @tonytlwu 

**Fixed behavior:**
When you use the form with a date picker and rich text (only in this sequence) form doesn't submit in IE11 and Edge. Now it submits and shows the correct message.

ref Fliplet/fliplet-studio#4790

richTextForm-Submit-Fix

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -474,7 +474,7 @@ Fliplet.Widget.instance('form-builder', function(data) {
           /**
            * When we try to submit a form in Edge or IE11 and use components date picker and rich text
            * (only in this sequence) we could saw that rich text textarea become empty but there was no 
-           * message that we successfully submited the form. That was because Vue wasn't updating view.
+           * message that we successfully submitted the form. That was because Vue wasn't updating view.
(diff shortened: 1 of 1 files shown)


js/libs/form.js
```diff
@@ -471,6 +471,13 @@ Fliplet.Widget.instance('form-builder', function(data) {
           $vm.isSent = true;
           $vm.isSending = false;
           $vm.reset(false);
+          /**
+           * When we try to submit a form in Edge or IE11 and use components date picker and rich text
+           * (only in this sequence) we could saw that rich text textarea become empty but there was no 
+           * message that we successfully submited the form. That was because Vue wasn't updating view.
```
squallstar: submitted*

tonytlwu:
```suggestion
           * message that we successfully submitted the form. That was because Vue wasn't updating view.
```

YaroslavOvdii:
@tonytlwu 
Updated file as suggested.

PR #221, opened 2019-08-07
fixes for file field when dealing with strings

squallstar:
ref https://github.com/Fliplet/fliplet-studio/issues/4795

### DIFF
--- a/js/components/file.js
+++ b/js/components/file.js
@@ -57,22 +57,31 @@ Fliplet.FormBuilder.field('file', {
     },
     onReset: function() {
       var $vm = this;
-      
+
       $vm.value = [];
       $vm.selectedFileName = '';
-      
+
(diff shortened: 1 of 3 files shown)


js/components/file.js
```diff
@@ -57,22 +57,31 @@ Fliplet.FormBuilder.field('file', {
...
       var $vm = this;
-      
+
       $vm.value = [];
       $vm.selectedFileName = '';
-      
+
       $vm.$emit('_input', $vm.name, $vm.value);
     },
+    ensureValue: function () {
```
tonytlwu: What does it ensure the value to be?

You could either call it `validateValue()`, which is what it's doing.

squallstar: Agreed and renamed 👍

PR #223, opened 2019-08-08
Refactor for hooks

squallstar:
I've rewritten https://github.com/Fliplet/fliplet-widget-form-builder/pull/222 after deciding to go for a different more robust approach

### DIFF
--- a/js/libs/builder.js
+++ b/js/libs/builder.js
@@ -435,7 +435,8 @@ var app = new Vue({
                 type: 'email',
                 runOn: ['update'],
                 widgetInstanceId: widgetId,
-                payload: payload
+                payload: payload,
+                triggers: [widgetUuid]
               });
             }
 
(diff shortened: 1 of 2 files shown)


js/libs/builder.js
```diff
@@ -528,7 +543,6 @@ var app = new Vue({
 
           ds.hooks.push({
             widgetInstanceId: widgetId,
-            widgetInstanceUuid: widgetUuid,
```
tonytlwu: I don't fully understand the whole logic, but just by pattern matching the rest of the PR changes: does this line not need a `source` attribute here, because you've removed all other `widgetInstanceUuid`s with `source`?

squallstar: @tonytlwu I originally had it but these hooks are the `beforeSave, beforeQuery` operations (e.g. password hashing), I think they shouldn't be affected in the same way that emails are. Do you agree?

PR #232, opened 2019-08-22
Added tinymce destroyed method

MaksymShokin:
@tonytlwu @squallstar 

## Issue
Fliplet/fliplet-studio#4827

## Description
Added destroyed method for TinyMCE after the user leaves the settings tab. Now bug should not occur.

## Screenshots/screencasts
tinymce demo 1

## Backward compatibility

### DIFF
--- a/js/libs/builder.js
+++ b/js/libs/builder.js
@@ -800,6 +800,8 @@ var app = new Vue({
       if (value === 'settings') {
         $vm.setupCodeEditor();
         changeSelectText();
+      } else {
+        tinymce.remove();
       }
     },
     'settings.dataStore': function(value) {


js/libs/builder.js
```diff
@@ -800,6 +800,8 @@ var app = new Vue({
       if (value === 'settings') {
         $vm.setupCodeEditor();
         changeSelectText();
+      } else {
+        tinymce.remove();
```
squallstar: Wouldn't it be better to basically only initialise TinyMCE once? Let me know if you agree or removing it is cleaner based on the rest of the code.

MaksymShokin: I tried to initialize TinyMCE once but in that case bug still occurs.

PR #237, opened 2019-10-11
Improve saving of user progress as the user types

MaksymShokin:
@tonytlwu @squallstar 

## Issue
Fliplet/fliplet-studio#1686

## Description
Added function that saves progress on input.

## Screenshots/screencasts
save demo

## Backward compatibility

## Notes
If the user changes screen too fast after typing (approximately less then 1s), it might not save. Apparently there is nothing we can do in that case.


js/libs/core.js
```diff
@@ -86,7 +86,11 @@ Fliplet.FormBuilder = (function() {
             $($vm.$el).removeClass('has-error');
           }
         }
-      }
+      };
+
+      component.methods.saveProgress = function($event) {
+        this.$emit('_input', this.name, $event.target.value);
```
tonytlwu: @squallstar This isn't debounced, while the user could be typing very quickly, firing this many times. Do you see a problem with it?

js/libs/core.js
```diff
@@ -86,7 +86,11 @@ Fliplet.FormBuilder = (function() {
             $($vm.$el).removeClass('has-error');
           }
         }
-      }
+      };
+
+      component.methods.saveProgress = function($event) {
```
squallstar: Please rename this method to `onInput`

js/libs/core.js
squallstar: I agree with debouncing this to something like 200ms. @tonytlwu confirm a good debounce time in regards to average typing speed :-)

templates/components/email.build.hbs
```diff
@@ -3,9 +3,10 @@
   class="form-control"
   v-model.trim.lazy="value"
   v-on:blur="updateValue()"
+  v-on:input="saveProgress"
```
squallstar: `v-on:input="onInput()"`

js/libs/core.js
MaksymShokin: Fixed

js/libs/core.js
MaksymShokin: Fixed

templates/components/email.build.hbs
MaksymShokin: Fixed

PR #238, opened 2019-10-25
Support for reading list of options on a field and getting the field vue instance

squallstar:
```js
// get
var options = form.field('foo').options()

// set 
form.field('foo').options([1,2,3])

// reference $vm instance (follows same pattern as form.instance)
form.field('foo').instance
```

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -726,6 +726,10 @@ Fliplet.Widget.instance('form-builder', function(data) {
                 }
               },
               options: function (values) {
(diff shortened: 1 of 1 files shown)


tonytlwu:
@squallstar Don't we already have the reading of options from a field? https://github.com/Fliplet/fliplet-studio/issues/2520 says it was already implemented.

squallstar:
@tonytlwu `get` wasn't working though.

PR #241, opened 2019-11-05
Removes empty data from retrieved entry before loading to avoid errors with form fields

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/5208

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -475,9 +475,9 @@ Fliplet.Widget.instance('form-builder', function(data) {
           $vm.reset(false);
           /**
            * When we try to submit a form in Edge or IE11 and use components date picker and rich text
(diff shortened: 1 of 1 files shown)


js/libs/form.js
```diff
@@ -519,6 +519,11 @@ Fliplet.Widget.instance('form-builder', function(data) {
               record = { data: record };
             }
 
+            record.data = _.omitBy(record.data, function (value) {
+              return _.isNil(value)
+                || (_.isObject(value) && _.isEmpty(value))
+                || _.isString(value) && !value.length;
```
squallstar: shouldn't the last line be all in a single parenthesis?

tonytlwu: It should, indeed

PR #245, opened 2019-11-08
If user changes options using setting code, previously selected value will be cleared if they do not correspont to new options

MaksymShokin:
@tonytlwu @squallstar @sofiiakvasnevska 

## Issue
Fliplet/fliplet-studio#5200

## Description
Added check to values when user setting options using code.

## Screenshots/screencasts
set options demo

## Backward compatibility

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -757,6 +757,19 @@ Fliplet.Widget.instance('form-builder', function(data) {
                   return { id: option };
                 });
 
+                if (field.value.length && field._type === 'flCheckbox') {
+                  var selectedValues = _.difference(field.value, values);
+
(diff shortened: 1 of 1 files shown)


js/libs/form.js
```diff
@@ -757,6 +757,19 @@ Fliplet.Widget.instance('form-builder', function(data) {
                   return { id: option };
                 });
 
+                if (field.value.length && field._type === 'flCheckbox') {
```
squallstar: let's nest the two blocks only when there's a value and use a switch case:

```
if (field.value) {
  switch (field._type) {
    case 'flCheckbox'
  }
}
```

MaksymShokin: @squallstar fixed

PR #249, opened 2019-11-13
Entering value for checkbox in string form and removing dublicates now works both for val and set methods

MaksymShokin:
@sofiiakvasnevska

## Issue
Fliplet/fliplet-studio#5222
Fliplet/fliplet-studio#5227

## Description
If the value set for checkbox it converts the value to array if it is a string and removes duplicates if it is an array. Works for both val and set methods.

## Screenshots/screencasts
set options demo

## Backward compatibility


js/libs/form.js
```diff
@@ -610,6 +610,18 @@ Fliplet.Widget.instance('form-builder', function(data) {
           $form.$forceUpdate();
         }, 10);
 
+        function validateCheckboxValue(value) {
+          if (typeof value === 'string') {
+            value = value.split(',');
```
squallstar: also add a `.map(function (value) { return value.trim(); });`, @tonytlwu do you agree?

tonytlwu: I agree. Also, is it likely that the value contains `,` as well and might be wrapped in `""`?

MaksymShokin: @tonytlwu 
In case we pass strings separated by commas only first sting will get selected.
picturemessage_xchd0qgd vpj

In case we pass one string separated inside by commas we select all values.
picturemessage_ziqcglde qm1

tonytlwu: What if the options for the field were:

- Option 1
- Option 2, with comma
- Option 3

tonytlwu: Also, why are we calling .val() with multiple parameters?

I can imagine calling `.val(['Option 1', 'Option 2', 'Option 3'])`, not `.val('Option 1', 'Option 2', 'Option 3')`.

MaksymShokin: @tonytlwu If the user will have a comma inside field option then it would not work because we use comma as a separator.  Do you have any ideas on how we should process it?

tonytlwu: We have a pretty extensively tested function for splitting a string by commas, while ignoring the commas that are wrapped in `"  "`s.

See https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/aa606369efcba8cd04df0905a8895f5b8269f361/js/utils.js#L130

MaksymShokin: @tonytlwu The function you provided would work in the case if the user wraps option in double-quotes. 
picturemessage_wrmjmue4 dbn

I assume that user would set value in the following form .val('Option 1, Option 2,, Option 3') in that case function would still fail.

tonytlwu: I was commenting on the way that we're using a "split by comma" operation, and that if we're doing this, then we should consider what happens if a comma was meant to be there.

I haven't looked at the wider concept of what this PR is trying to do. For me, a `.val('Foo bar')` or `.set('Foo bar')` would both do the following when it comes to checkboxes:

1. Check if the parameter is an array. If not put it into an array of 1 element.
1. Remove any values in the parameter that isn't an available option for the field
1. Remove duplicates
1. Makes sure that when the field is submitted, the values appear in the order as configured in the available options -- this might be done during submission.

@squallstar Do you have anything to add? You should review and approve the solution is correct.

squallstar: @tonytlwu I agree with your suggested solution 👍

js/libs/form.js
```diff
@@ -610,6 +610,20 @@ Fliplet.Widget.instance('form-builder', function(data) {
           $form.$forceUpdate();
         }, 10);
 
+        function validateCheckboxValue(value) {
+          if (typeof value === 'string') {
```
tonytlwu: I don't understand why we're checking if the value is a string. We just need to check if it's an array or not.

```suggestion
          if (Array.isArray(value)) {
            value = [value];
          }
```

What's the purpose of splitting by `,` here, based on the recommended flow ([link](https://github.com/Fliplet/fliplet-widget-form-builder/pull/249#discussion_r350711438))?

js/libs/form.js
MaksymShokin: fixes

tonytlwu: @MaksymShokin I'm not sure if this is resolved.

The first line in this function still starts with:

```js
if (!_.isArray(value)) {
  value = value.split(',');
}
```

This has no correlation to the first step I mentioned above:

> 1. Check if the parameter is an array. If not **put it into an array of 1 element**.

If I'm wrong, can you explain how your code relates to what we've outlined ☝️ ?

tonytlwu: @MaksymShokin I'm not sure if this is resolved.

The first line in this function still starts with:

```js
if (!_.isArray(value)) {
  value = value.split(',');
}
```

This has no correlation to the first step I mentioned above:

> 1. Check if the parameter is an array. If not **put it into an array of 1 element**.

If I'm wrong, can you explain how your code relates to what we've outlined ☝️ ?

MaksymShokin:
@tonytlwu 
Splitting by commas would be beneficial if the user forgets to put parameter into the array and would pass parameters like this: 
  form.field('field-1').val('foo, 111, 222, 333');
});

In case user will simply pass : 
  form.field('field-1').val(' 'string without comma' );
}); 
It will put this into an array of one element // ['string without comma'] .

tonytlwu:
@MaksymShokin 

> [...] beneficial if the user forgets to put parameter into the array

This is out of scope.

js/libs/form.js
```diff
@@ -655,6 +655,34 @@ Fliplet.Widget.instance('form-builder', function(data) {
           $vm.saveProgress();
         }, 10);
 
+        function validateCheckboxValue(value, options) {
+          value = _.isArray(value) ? value : [value];
+
+          value = value.map(function(elem) {
+            return elem.trim();
```
tonytlwu: I don't think this part is necessary. I vaguely remembered @squallstar suggested using `.trim()` because there was a `.split(",")` operation happening and we want to make sure the values are trimmed. We're not doing that any more so this part shouldn't be necessary.

@squallstar Thoughts?

js/libs/form.js
```diff
@@ -655,6 +655,34 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
+          });
+
+          if (options.length) {
+            var valueInOptions = [];
```
tonytlwu: Why are we setting up an empty array and then one-by-one pushing values into the array?

We should be able to use:

```suggestion
            var valueInOptions = _.filter(values, fn);
```

js/libs/form.js
```diff
@@ -655,6 +655,34 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
+
+            value.forEach(function(elem) {
+              if (_.find(options, { label: elem })) {
```
tonytlwu: I remember users can set values to a checkbox that is different from the label, e.g.

```
Option 1 <1>
```

...would produce an option that has the label `Option 1` and value `1`. In this case, we should be matching by the value, not by the label.

This line here suggests that we're matching by label.

js/libs/form.js
squallstar: @tonytlwu if we're not using `split` then we shouldn't `trim`, that's correct.

js/libs/form.js
squallstar: @MaksymShokin ^

js/libs/form.js
MaksymShokin: fixed

js/libs/form.js
MaksymShokin: fixed

js/libs/form.js
MaksymShokin: fixed

squallstar:
FYI this PR broke our forms as using `form.load` would break form fields such as title and paragraph. I have fixed this here https://github.com/Fliplet/fliplet-widget-form-builder/pull/262

MaksymShokin:
@squallstar Thanks for the information!

PR #252, opened 2019-11-21
Added validation for data that we take from the data source

YaroslavOvdii:
@sofiiakvasnevska 
## Issue
https://github.com/Fliplet/fliplet-studio/issues/5287
https://github.com/Fliplet/fliplet-studio/issues/5280
https://github.com/Fliplet/fliplet-studio/issues/5286
https://github.com/Fliplet/fliplet-studio/issues/5285
https://github.com/Fliplet/fliplet-studio/issues/5282
https://github.com/Fliplet/fliplet-studio/issues/5274
https://github.com/Fliplet/fliplet-studio/issues/5275
https://github.com/Fliplet/fliplet-studio/issues/5276

## Description
Added validation for data that we take from the data source

## Screenshots/screencasts
Rich text
Rich-text

Upload file
File

Date picker
Date

Time picker
Time

Checkbox
checkbox

Radio button
Radio

Star rating
Star

## Backward compatibility

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -163,9 +163,8 @@ Fliplet.Widget.instance('form-builder', function(data) {
                 var inOptions = [];
 
                 entry.data[field.name].forEach(function(element) {
-                  var inOption = _.find(field.options, function(option) {
-                    return option.label === element
-                  });
+                  var inOption = _.find(field.options, { label: element });
+
                   if (inOption) {
                     inOptions.push(inOption);
                   }
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -114,31 +114,67 @@ Fliplet.Widget.instance('form-builder', function(data) {
 
     if (fields.length && (data.saveProgress && typeof progress === 'object') || entry) {
       fields.forEach(function(field) {
-        if (entry && entry.data && typeof entry.data[field.name] !== 'undefined' && field.populateOnUpdate !== false) {
-          if (field._type === "flDate") {
-            if (Fliplet.Env.get('platform') === 'web') {
-              field.value = moment(entry.data[field.name]).format('DD MMMM YYYY');
-            } else {
-              field.value = moment(entry.data[field.name]).format('YYYY-MM-DD');
-            }
-          } else if (field._type === "flCheckbox" && !Array.isArray(entry.data[field.name])) {
-            field.value = [];
-          } else if (field._type === 'flImage') {
-            var img = entry.data[field.name];
-            field.value = [];
-            if (Array.isArray(img)) {
-              field.value = img;
-            } else if (typeof img === 'string') {
-              field.value.push(img);
-            }
+        if (entry && entry.data && field.populateOnUpdate !== false) {
+          switch (field._type) {
+            case 'flDate':
+              var regex = /([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))/;
+              field.value = regex.exec(entry.data[field.name]) ? entry.data[field.name] : moment().get().format('YYYY-MM-DD');
+              break;
 
-            field.value = field.value.map(function (url) {
-              return Fliplet.Media.authenticate(url);
-            });
-          } else if (!field._submit && typeof field._submit !== 'undefined') {
-            field.value = field.value;
-          } else {
-            field.value = entry.data[field.name];
+            case 'flImage':
+            case 'flFile':
+              var img = entry.data[field.name];
+              field.value = [];
+
+              if (Array.isArray(img)) {
+                field.value = img;
+              } else if (typeof img === 'string') {
+                field.value.push(img);
+              }
+    
+              field.value = field.value.map(function(url) {
+                return Fliplet.Media.authenticate(url);
+              });
+              break;
+
+            case (!field._submit && typeof field._submit !== 'undefined' || typeof entry.data[field.name] === 'undefined'):
+              field.value = field.value;
+              break;
+
+            case 'flTime':
+              var regexp = /^([0-1]?[0-9]|2[0-4]):([0-5][0-9])(:[0-5][0-9])?$/;
+
+              if (regexp.exec(entry.data[field.name])) {
+                field.value = entry.data[field.name];
+              } else {
+                field.value = moment().get().format('HH:mm');
+              }
+              break;
+
+            case 'flCheckbox':
+            case 'flRadio':
+            case 'flStarRating':
+              if (Array.isArray(entry.data[field.name]) && entry.data[field.name].length > 0) {
+                var inOptions = [];
+
+                entry.data[field.name].forEach(function(element) {
+                  var inOption = _.find(field.options, function(option) {
+                    return option.label === element
+                  });
+                  if (inOption) {
+                    inOptions.push(inOption);
+                  }
+                });
+                
+                field.value = inOptions.length ? _.uniqWith(entry.data[field.name], _.isEqual) : [];
+              } else {
+                field.value = [];
+              }
+              break;
+
+            default:
+              field.value = entry.data[field.name];
+              break;
           }
 
           return field.value;
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -172,18 +172,6 @@ Fliplet.Widget.instance('form-builder', function(data) {
               }
               break;
 
-            case (!field._submit && typeof field._submit !== 'undefined' || typeof entry.data[field.name] === 'undefined'):
-              /**
-               * In case if field._submit is not undefined or we received undefined entry.data[field.name]
-               * for example a user forgot to add any value in his code.
-               * In these cases, we do not change a field value
-               * We moved this `typeof entry.data[field.name] === 'undefined' ` condition to this position 
-               * because when we receive an undefined value in the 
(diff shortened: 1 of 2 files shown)


js/libs/form.js
```diff
@@ -114,31 +114,67 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
-            if (Array.isArray(img)) {
-              field.value = img;
-            } else if (typeof img === 'string') {
-              field.value.push(img);
-            }
+        if (entry && entry.data && field.populateOnUpdate !== false) {
+          switch (field._type) {
+            case 'flDate':
+              var regex = /([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))/;
+              field.value = regex.exec(entry.data[field.name]) ? entry.data[field.name] : moment().get().format('YYYY-MM-DD');
```
squallstar: shouldn't we parse the `entry.data[field.name]` with moment?

js/libs/form.js
```diff
@@ -114,31 +114,67 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
-            field.value = [];
-          } else if (field._type === 'flImage') {
-            var img = entry.data[field.name];
-            field.value = [];
-            if (Array.isArray(img)) {
-              field.value = img;
-            } else if (typeof img === 'string') {
-              field.value.push(img);
-            }
+        if (entry && entry.data && field.populateOnUpdate !== false) {
```
squallstar: any reason on why `typeof entry.data[field.name] !== 'undefined'` has been removed?

js/libs/form.js
```diff
@@ -114,31 +114,67 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
+              break;
 
-            field.value = field.value.map(function (url) {
-              return Fliplet.Media.authenticate(url);
-            });
-          } else if (!field._submit && typeof field._submit !== 'undefined') {
-            field.value = field.value;
-          } else {
-            field.value = entry.data[field.name];
+            case (!field._submit && typeof field._submit !== 'undefined' || typeof entry.data[field.name] === 'undefined'):
```
squallstar: move this as last condition and add a comment explaining what it does

js/libs/form.js
```diff
@@ -114,31 +114,67 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
+              field.value = field.value;
```
squallstar: this feels useless, can you explain?

js/libs/form.js
```diff
@@ -114,31 +114,67 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
+              break;
+
+            case 'flTime':
+              var regexp = /^([0-1]?[0-9]|2[0-4]):([0-5][0-9])(:[0-5][0-9])?$/;
+
+              if (regexp.exec(entry.data[field.name])) {
+                field.value = entry.data[field.name];
+              } else {
+                field.value = moment().get().format('HH:mm');
```
squallstar: shouldn't moment parse `entry.data[field.name]`?

js/libs/form.js
```diff
@@ -114,31 +114,67 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
+
+            case 'flCheckbox':
+            case 'flRadio':
+            case 'flStarRating':
+              if (Array.isArray(entry.data[field.name]) && entry.data[field.name].length > 0) {
+                var inOptions = [];
+
+                entry.data[field.name].forEach(function(element) {
+                  var inOption = _.find(field.options, function(option) {
+                    return option.label === element
```
squallstar: missing semicolon

js/libs/form.js
```diff
@@ -114,31 +114,67 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
+              break;
+
+            case 'flCheckbox':
+            case 'flRadio':
+            case 'flStarRating':
+              if (Array.isArray(entry.data[field.name]) && entry.data[field.name].length > 0) {
+                var inOptions = [];
+
+                entry.data[field.name].forEach(function(element) {
+                  var inOption = _.find(field.options, function(option) {
```
squallstar: `_.find(field.options, { label: element })` should work too

js/libs/form.js
YaroslavOvdii: @squallstar 
If we will try to parse threw the moment `entry.data[field.name]` when we get a 0 or 1 it will see this value as valid and give as this [result](https://jsfiddle.net/YaOvdiy/Lh35qgyz/6/).
Therefore regexp is safer in this situation.

js/libs/form.js
YaroslavOvdii: @squallstar I wasn't adding this. It already was in the file. As far as I understand this line saves field.value in the case when the submit button isn't available and when new data is `undefined`.

js/libs/form.js
squallstar: @YaroslavOvdii an ISO date should be considered valid, e.g. `2019-11-22T14:45:58+01:00` should be parsed with moment as it's a valid date. Likewise, `2019` alone is also valid if you try with moment. It gives a warning but the date is parsed correctly to `2019-01-01`.

Given the value exists, I would prefer parsing it. @tonytlwu what do you think?

js/libs/form.js
squallstar: @YaroslavOvdii the original code is `!field._submit && typeof field._submit !== 'undefined'`. Your change breaks it because you have `a && b || c`. Have you tested this?

js/libs/form.js
YaroslavOvdii: @squallstar it wasn't removed just moved to the 140th line.
I moved it because of [this](https://github.com/Fliplet/fliplet-studio/issues/5275) task. There we needed that when undefined was sent as data for the date field we should set the date field to the current date. And when we had that condition at the beginning we were passing this data. The same was needed for [this](https://github.com/Fliplet/fliplet-studio/issues/5276) task as well.

js/libs/form.js
YaroslavOvdii: @squallstar same reason [this](https://github.com/Fliplet/fliplet-widget-form-builder/pull/252#discussion_r349602010).

js/libs/form.js
YaroslavOvdii: @squallstar Yes, I have tested it. It worked.
The reason why I was adding or condition here is because of in this case `typeof entry.data[field.name] === 'undefined'` we were skipping all other if condition and just returning `field.value` which is equal for this condition as I can see. 
Was it used for another purpose?

js/libs/form.js
squallstar: @tonytlwu @YaroslavOvdii bump ^

tonytlwu: 1. I'm happy with supporting an ISO date string, and extracting the date portion of it as the value, e.g. `2019-11-22T14:45:58+01:00` is parsed as `2019-11-22`. Saving the value through the form will then transform it to `2019-11-22`, essentially dropping the time.
1. I don't think `2019` should be parsed into a date. We should just show today's date. I'm happy to discuss additional requirements for how to cleverly interpret other date string patterns.

YaroslavOvdii: @tonytlwu @squallstar 
I agreed that supporting an ISO date string is needed.
If I may suggest, what if we will extend regexp and look for ISO format string as well with the date string.
Therefore we shall do 2 regexp for the date field.
We will check if the string is in the date string format or the ISO string format.

tonytlwu: @YaroslavOvdii Yes, I agree with checking for something like `2019-11-22T14:45:58+01:00` as well as `2019-11-22`.

js/libs/form.js
YaroslavOvdii: @squallstar Fixed

js/libs/form.js
YaroslavOvdii: @squallstar Fixed

js/libs/form.js
YaroslavOvdii: @tonytlwu @squallstar Added checking for the ISO date format in the date block.

tonytlwu: Happy with this part of the PR.

js/libs/form.js
```diff
@@ -176,6 +172,18 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
+            case (!field._submit && typeof field._submit !== 'undefined' || typeof entry.data[field.name] === 'undefined'):
+              /**
+               * In case if field._submit is not undefined or we received undefined entry.data[field.name]
+               * for example a user forgot to add any value in his code.
+               * In these cases, we do not change a field value
+               * We moved this `typeof entry.data[field.name] === 'undefined' ` condition to this position 
+               * because when we receive an undefined value in the 
+               * date or time input we should set them to the current date or time.
+              */
+              field.value = field.value;
```
tonytlwu: I don't understand why we have this `case` at all if all it does is assign a variable to itself. It looks like it doesn't do anything at all, thus the `case` is unnecessary?

YaroslavOvdii: @tonytlwu  
This case was initially here, as far as I understand it I described in the comment above it.
But when I tried to removed this case nothing was changed.
Should I remove it from the code?

squallstar: I agree with the analysis, let's remove this unnecessary block.

PR #259, opened 2019-12-16
Invalid options are removed when setting dropdown field options

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5448

## Description
The invalid options are removed before the options are set.

## Screenshots/screencasts
dropdown options demo

## Backward compatibility

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -136,7 +136,7 @@ Fliplet.Widget.instance('form-builder', function(data) {
               } else if (typeof img === 'string') {
                 field.value.push(img);
               }
-    
+
               field.value = field.value.map(function(url) {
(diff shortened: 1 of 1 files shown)


js/libs/form.js
```diff
@@ -822,6 +822,12 @@ Fliplet.Widget.instance('form-builder', function(data) {
                   throw new Error('Options must be an array');
                 }
 
+                if (field._type === 'flSelect') {
+                  _.remove(values, function(val) {
+                    return val === null  || val === undefined || val.trim() === '';
```
squallstar: there's a double space here which can be removed

MaksymShokin: fixed

PR #266, opened 2020-01-21
On password and email fields autocomplete turned off

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5494

## Description
All other methods like autocomplete='off' don't work on Chrome.

## Screenshots/screencasts
autocomplete-off

## Backward compatibility


squallstar:
@MaksymShokin have you confirmed whether this works fine when programmatically setting form values for those fields and if it works on all browsers including native platforms (iOS and Android) under all conditions, e.g. moving between fields by either tapping or tabbing between fields?

MaksymShokin:
@squallstar 
Apparently it does not work perfectly on native devices (Fliplet Viewer). I am not sure how to fix it then. Chrome ignores any attributes like autocomplete='off'.

MaksymShokin:
@squallstar I found what the issue is. Adding readonly will not work. I will close this PR and try to find another solution.

PR #267, opened 2020-01-29
Autocomplete disabled

MaksymShokin:
@sofiiakvasnevska

Issue
Fliplet/fliplet-studio#5494

Description
Apparently this method works fine, although sometimes Chrome anyway autofills password. Maybe we should wait for Inna to test it before making PR in other components.

Screenshots/screencasts
autocomplete-off

Backward compatibility


templates/components/password.build.hbs
```diff
@@ -2,7 +2,7 @@
   type="password"
   class="form-control"
   :readonly="autogenerate"
-  autocomplete="off"
+  autocomplete="new-password"
```
squallstar: why `new-password`? If it needs to be a random field name then you can generate it with `Fliplet.guid()`

MaksymShokin: This is the only way it works.

squallstar: @maksim-an-upplabs can you elaborate? post relevant documentation to support the argument?

squallstar: @MaksymShokin can you elaborate? post relevant documentation to support the argument?

MaksymShokin: @squallstar 
After I tried other methods I found this article https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion.
Below it has section called "Preventing autofilling with autocomplete="new-password"."

PR #273, opened 2020-02-21
Users now cannot save empty button label

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5485

## Description
This fix should solve the problem mentioned in the issue but also I decided to fix the possible issue that user can save the empty button label.

## Screenshots/screencasts
buttons fix

## Backward compatibility


tonytlwu:
@squallstar Could you take a look at this please?

PR #277, opened 2020-02-24
Setting default checkboxes values now works correctly

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5417

## Description
The user is not able to select multiple checkboxes default values.

## Screenshots/screencasts
checkboxes

## Backward compatibility


squallstar:
where it the code that splits the value by newline whenever the input changes?

MaksymShokin:
@squallstar 
I did not understand the question. Value splits in line 61 in checkbox.js.

squallstar:
@MaksymShokin that line seem responsible for setting an existing value on the form. What about when the user is actually typing more default values here https://github.com/Fliplet/fliplet-widget-form-builder/pull/277/files#diff-363e1f42405d0c0c5f18b227721e90fcR3

Have you confirmed whether new values are correctly saved in the settings? I would have expected that input to save the input value as an array instead of a string.

MaksymShokin:
@squallstar 
The logic behind it is that if the user sets default value (though textarea you've highlighted), it is technically a string, therefore it will be splitted to an array. If the user sets value though code or has selected any value, split code would not run. I checked and the value saves correctly

PR #282, opened 2020-03-04
Make validation work right away as user start to input something into the URL field

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5405

## Description
Make validation work right away as user start to input something into the URL field

## Screenshots/screencasts
https://share.getcloudapp.com/kpuYw4Ge

## Backward compatibility

## Notes
Changed event from blur to focus so that validation rule starts to work right away but not waiting element to be blurred.

### DIFF
--- a/js/build.templates.js
+++ b/js/build.templates.js
@@ -87,7 +87,7 @@ this["Fliplet"]["Widget"]["Templates"]["templates.components.title"] = Handlebar
 },"useData":true});
 
 this["Fliplet"]["Widget"]["Templates"]["templates.components.url"] = Handlebars.template({"compiler":[7,">= 4.0.0"],"main":function(container,depth0,helpers,partials,data) {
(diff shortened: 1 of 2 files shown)


tonytlwu:
@YaroslavOvdii I don't think the UX from your screencast works. Your video showed that when the user is typing (although I can't tell clearly because of the autocomplete dropdown) the error is appearing. This means that if I'm trying to type `https://fliplet.com` and have typed the first two `ht` characters, I'm going to get an invalid URL warning. This is too soon.

1. The field should be validated on `blur`, and users be warned of errors where relevant
1. The field should be validated on `submit`, and users be warned of errors where relevant

YaroslavOvdii:
@tonytlwu do I get it right that if a user will try to submit the form with the field which didn't pass the validation we should deny submit and show to a user the field with error?
There are no such behavior for now.

tonytlwu:
@YaroslavOvdii Yes, if a user tries to submit the form with at least one field that doesn't pass the validation, then the submission should be denied.

As far as I see, that's exactly what was reported in [the original issue](https://github.com/Fliplet/fliplet-studio/issues/5405):

> 1. Login > Create an app
> 2. Drop the Form component onto a screen
> 3. Add URL field > Save
> 4. Proceed to the Preview
> 5. Enter invalid data in URL input filed
> 6. Pay attention that 'The input is not a valid URL' or 'Please enter a URL' messages do not appear and it is possible to submit the form.

YaroslavOvdii:
@tonytlwu

> @YaroslavOvdii Yes, if a user tries to submit the form with at least one field that doesn't pass the validation, then the submission should be denied.

 I'll do a new PR with this solution and close this PR.

PR #283, opened 2020-03-05
Clearing form fields when clicking on the "Clear" button

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5514

## Description
Clearing form fields when clicking on the "Clear" button

## Screenshots/screencasts
https://share.getcloudapp.com/7KuRpbdO

## Backward compatibility

## Notes
As I tested I couldn't see any effect on other parts of the app. Could you please verify this.

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -268,7 +268,7 @@ Fliplet.Widget.instance('form-builder', function(data) {
             value = value.slice(0);
           }
 
-          field.value = value;
+          field.value = '';
           $vm.triggerChange(field.name, field.value);
         });
 


js/libs/form.js
```diff
@@ -268,7 +268,7 @@ Fliplet.Widget.instance('form-builder', function(data) {
             value = value.slice(0);
           }
 
-          field.value = value;
+          field.value = '';
```
tonytlwu: @squallstar Could you review this one? It seems correct to me, but it's also extremely suspicious why this bug would have been here and perhaps you coded it for a reason?

```js
field.value = value;
```

...makes it looks like nothing gets reset at all, which would have seemed obvious. I'll let you review and confirm this is a suitable fix.

squallstar: `field.value = value;` is there for a reason: it changes the field value with the actual value taken from the form configuration, so that the default value is applied.

@YaroslavOvdii did you test your patch with
- all field types having a default value which is not an empty string? When pressing reset the field value should reset to such value, not to an empty string as you coded
- what happens if the form progress is restored and then reset is pressed

YaroslavOvdii: @squallstar 
To clarify we should not clear the value of the field if there is a default value was set, but just return the field to it default value?

squallstar: @YaroslavOvdii yes, that's correct. If I have set my text input to have a default value of "Foo", then resetting the form should bring the field value back to "Foo", not an empty string. @tonytlwu can you confirm too this is the behaviour we want?

tonytlwu: @squallstar That's correct. If a default value is provided, it should reset the value back to what the default was. The button's action is more like a "Reset" than a "Clear".

YaroslavOvdii: @tonytlwu @squallstar 
I got it. Then I'll make a new commit with the correct solution.

YaroslavOvdii: @tonytlwu 
Does it mean that we should revert this [PR](https://github.com/Fliplet/fliplet-widget-form-builder/pull/280)?

tonytlwu: Yes I think so. See https://github.com/Fliplet/fliplet-studio/issues/5514#issuecomment-600000640 for how this should behave.

YaroslavOvdii:
I'm closing this PR because we do not need these changes anymore.

PR #287, opened 2020-03-11
Organization form template description is displayed correctly

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5412

## Description
We clear the default description only when selecting the template. Afterwards if user sets new description it will be saved correctly.

## Screenshots/screencasts
form-template

## Backward compatibility


tonytlwu:
@squallstar This is for an issue that you (and @hcarneiro ) were giving advice on. I'll let you review.

PR #293, opened 2020-03-25
Checkbox and rich text values will be restored on clear button press

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6087

## Description
Checkbox and rich text values will be restored on clear button press

## Screenshots/screencasts
https://share.getcloudapp.com/OAurZAyk

## Backward compatibility

### DIFF
--- a/js/components/checkbox.js
+++ b/js/components/checkbox.js
@@ -6,10 +6,6 @@ Fliplet.FormBuilder.field('checkbox', {
       type: Array,
       default: []
     },
-    defaultValue: {
-      type: String,
-      default: ''
-    },
     options: {
       type: Array,
       default: [
@@ -61,12 +57,11 @@ Fliplet.FormBuilder.field('checkbox', {
     }
   },
   created: function () {
-    if (!!this.defaultValue) {
-      this.value = this.defaultValue.split(/\n/);
-      this.updateValue(this.name, this.value);
-    } else if (!Array.isArray(this.value)) {
-      this.value = [];
-      this.updateValue(this.name, this.value);
+    // We have removed defaultValue variable and set value intead
+    // Because default value wasn't saving in the global data variable ib that case and we couldn't restore default value
+    if (!Array.isArray(this.value)) {
+      this.value = this.value.split(/\n/);
+      this.updateValue();
     }
   }
 });
--- a/js/components/wysiwyg.js
+++ b/js/components/wysiwyg.js
@@ -42,11 +42,9 @@ Fliplet.FormBuilder.field('wysiwyg', {
     onReset: function () {
       if (this.editor) {
         try {
-          return this.editor.setContent('');
+          return this.editor.setContent(this.value);
         } catch (e) {}
       }
-
-      this.value = '';
     },
     placeholderLabel: function () {
       var placeholderText = this.editor.getElement().getAttribute("placeholder") || this.editor.settings.placeholder;
--- a/templates/configurations/checkbox.interface.hbs
+++ b/templates/configurations/checkbox.interface.hbs
@@ -1,4 +1,4 @@
 <div class="form-group">
   <label>Default values <small>(One per line)</small></label>
-  <textarea class="form-control" v-model.trim="defaultValue" placeholder="Default values" />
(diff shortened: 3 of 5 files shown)


js/components/wysiwyg.js
```diff
@@ -42,11 +42,9 @@ Fliplet.FormBuilder.field('wysiwyg', {
     onReset: function () {
       if (this.editor) {
         try {
-          return this.editor.setContent('');
+          return this.editor.setContent(this.value);
```
squallstar: isn't this the current value, not the initial?

YaroslavOvdii: No, at this point this.value was already reassigned to the initial value

js/components/checkbox.js
```diff
@@ -61,12 +57,11 @@ Fliplet.FormBuilder.field('checkbox', {
     }
   },
   created: function () {
-    if (!!this.defaultValue) {
-      this.value = this.defaultValue.split(/\n/);
-      this.updateValue(this.name, this.value);
-    } else if (!Array.isArray(this.value)) {
-      this.value = [];
-      this.updateValue(this.name, this.value);
+    // We have removed defaultValue variable and set value intead
```
squallstar: I don't understand why the comment is mentioning something that should be mentioned in the PR description.

There shouldn't be a comment here as the code is straight forward

js/components/checkbox.js
```diff
@@ -61,12 +57,11 @@ Fliplet.FormBuilder.field('checkbox', {
...
+    // Because default value wasn't saving in the global data variable ib that case and we couldn't restore default value
+    if (!Array.isArray(this.value)) {
+      this.value = this.value.split(/\n/);
+      this.updateValue();
```
squallstar: don't `updateValue()` need to run if the value is an array?

js/components/checkbox.js
```diff
@@ -6,10 +6,6 @@ Fliplet.FormBuilder.field('checkbox', {
       type: Array,
       default: []
     },
-    defaultValue: {
```
squallstar: What about existing widget instance settings using the default value?

templates/configurations/checkbox.interface.hbs
```diff
@@ -1,4 +1,4 @@
 <div class="form-group">
   <label>Default values <small>(One per line)</small></label>
-  <textarea class="form-control" v-model.trim="defaultValue" placeholder="Default values" />
+  <textarea class="form-control" v-model.trim="value" placeholder="Default values" />
```
squallstar: what about current users having configured the `defaultValue`?

js/components/checkbox.js
YaroslavOvdii: @squallstar 
The main reason for this change is that I wasn't able to find how we set data into the `data.fields` array. And the only way to add data there was to change `defaultValue` to the `value`. If you could guide me where I can change that we need to send a `defaultValue` for Checkboxes it would be the best fix in my opinion.

squallstar: @YaroslavOvdii can you elaborate a little more on this:

> _"The main reason for this change is that I wasn't able to find how we set data into the data.fields array"_

YaroslavOvdii: @squallstar at this [line](https://github.com/Fliplet/fliplet-widget-form-builder/blob/b038c61edfbdbd81053a4c727b94dc89fbb37bfc/js/libs/form.js#L264) we take old field value from the data object. As I have noticed I can write data there only if the default value will be attached to `this.value`. Is there any way we could write `this.defaultValue` there?
I wasn't able to find where this object is set.

squallstar: @YaroslavOvdii `data.fields` is the list of fields configured via the interface as saved in the widget instance settings. I am not understanding what the current problem with the checkboxes are in this situation, but I suppose the problem could be that you are assigning the `value` directly so it wouldn't be parsed by any logic in the `created()` method of the checkboxes (e.g. the logic that splits strings into an array). Is this the case?

YaroslavOvdii:
Made a new solution for this issue https://github.com/Fliplet/fliplet-widget-form-builder/pull/294

PR #305, opened 2020-05-26
[Fix] Changed date field configuration to use date picker

MaksymShokin:
@sofiiakvasnevska

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6368

## Description
Added moment formatting to 2 places (when saving this.value, and when passing value to date picker) this is necessary in order to correctly display date.

## Screencast
https://streamable.com/q15bif

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/components/date.js
```diff
@@ -1,4 +1,4 @@
-var DATE_FORMAT = 'YYYY-MM-DD';
+var DATE_FORMAT = 'MM-DD-YYYY';
```
squallstar, line 1: This is meant to be `dd/mm/yyyy` if you read the linked issue, right?

MaksymShokin: @squallstar 
Actually I thought that both 'MM-DD-YYYY' and 'DD-MM-YYYY' are fine because of this comment.
Capture

Should I remake it to 'DD-MM-YYYY' ?

squallstar: I will discuss with @tonytlwu in the morning to ensure we have considered all applicable scenarios for existing apps and clients in both regions.

MaksymShokin:
@squallstar 
Thanks for guidance on this task, now I understand how form builder connections work. Fixes CSS. 
https://streamable.com/vntmex

PR #306, opened 2020-06-02
[Feat] Use date picker on date field configuration

squallstar:
Functionality works fine but it needs some CSS fixes:

image

ref https://github.com/Fliplet/fliplet-studio/issues/6368#issuecomment-637331349

@MaksymShokin please take it from here and have a go at fixing the CSS.


squallstar:
Implemented in https://github.com/Fliplet/fliplet-widget-form-builder/pull/305#pullrequestreview-422547161

PR #308, opened 2020-06-16
[WIP] [Feat] Match the datepicker start of week with the user region 

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/6452


tonytlwu:
StackOverflow seems to suggest that the datepicker will automatically pick up the moment locale info if locale is correctly set. I'll need to test this out once we've implemented moment locales.

PR #310, opened 2020-07-01
[Feature] Form builder accessibility

MaksymShokin:
@sofiiakvasnevska

## Issue
Fliplet/fliplet-studio#6422

## Description
Added styles that will highlight the selected input.

## Screencast
https://streamable.com/i1j16p

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


tonytlwu:
I agree with what @squallstar said. `outline` pretty much supports the same values as `border` and it'll affect the page layout less.

MaksymShokin:
Fixed.

squallstar:
@MaksymShokin can you update the original video with the updated version showing the changes?

MaksymShokin:
@squallstar 
Here is a new look. I changed buttons focus from outline to changing text color because outline doesnot look good on buttons. 
https://streamable.com/jxl7oi

squallstar:
I think the problem with changing the button text is that such button colour can be customised via the app appearance settings. If I change the background colour to #333 then the label becomes invisible. 

@tonytlwu what do you think? I think we should stick to using the outline and optionally make the font-weight bold when the button is focused. Alternative, we can even try a slight scale up via the CSS transform (e.g. `scale(1.05, 1.05)`) but we don't use that visual language so I'm not that keen about exploring such option.

tonytlwu:
I think outline should be fine. I'm not sure why outline wouldn't look good. And when everything else is using outlines, having one UI not using outlines creates inconsistency.

MaksymShokin:
@tonytlwu @squallstar 
By not looking good I mean this. We cannot add corner radius to outline. Therefore it looks like this. 
Capture

squallstar:
What about using the browser's default:

- https://stackoverflow.com/questions/20609485/what-is-the-default-style-of-the-blue-focus-outline-in-chrome/51358900
- http://jsfiddle.net/CXUpm/916/

tonytlwu:
I see. Use `box-shadow` instead then.

MaksymShokin:
@tonytlwu @squallstar 
To make everything correctly with outline we need to add an extra function on the label, like this.
outline 1
outline 2

Maybe it would be simpler to use border instead of outline? Therefore we would not need to add extra function.

tonytlwu:
@MaksymShokin The `html` tag has `context-build` and `context-interface` classes that you can use, depending on if it's in an app preview or Studio interface.

tonytlwu:
We can't use `border` because UI would sometimes already have its own set of borders. We can't use `outline` because it doesn't follow the border radius set by the user.

User `box-shadow`.

You can add a new function, but do not use the jQuery selector with a `#` ID. You should be able to use the VueJS `v-el` markup to define and access a node.

MaksymShokin:
I create new PR for this task, too many conflicts

PR #314, opened 2020-07-16
[Feature] Added possibility to add a explanation to a field

MaksymShokin:
@sofiiakvasnevska

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6491

## Description
I added explanation field to all inputs except buttons, paragraph, title, and horizontal line.

## Screencast
form-exp

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


MaksymShokin:
@tonytlwu @squallstar 
Fixed
https://streamable.com/yy1tp4

templates/components/radio.build.hbs
```diff
@@ -1,4 +1,5 @@
 <template>
+   <small v-if="description" class="help-block description">\{{ description }}</small>
```
tonytlwu, line 2:
```suggestion
  <small v-if="description" class="help-block description">\{{ description }}</small>
```

MaksymShokin: Fixed

templates/components/date.build.hbs
```diff
@@ -1,3 +1,4 @@
+<small v-if="description" class="help-block description">\{{ description }}</small>
```
tonytlwu, line 1: I prefer to put this inside a block element, e.g. `div`. If you want to use `small`, put the `small` tag inside a block element.

This way, the text by default will form a block on top of the input fields and makes it easier to style and adjust layouts for if necessary. In contrary, adding a block container later in the future may cause more problem.

MaksymShokin: @tonytlwu 
In that case we do not even need small tag.

tonytlwu:
@MaksymShokin Can you share the latest screenshot for how this looks? The original screenshot shows the explanation text under the fields instead of above them.

MaksymShokin:
@tonytlwu 
https://streamable.com/yy1tp4

PR #316, opened 2020-07-22
[Feature] Added possibility for users to add hidden fields

MaksymShokin:
@sofiiakvasnevska

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6790

## Description
I have to create a separate template for interface to differentiate between interface and build.

## Screencast
https://streamable.com/hhzr4o

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/components/file.js
```diff
@@ -33,6 +33,10 @@ Fliplet.FormBuilder.field('file', {
     value: {
       type: Array,
       default: []
+    },
+    isPossibleToHide: {
```
tonytlwu, line 37:
```suggestion
    canHide: {
```

MaksymShokin: Fixed

PR #318, opened 2020-07-23
[Fix] Removed warning

MaksymShokin:
@sofiiakvasnevska

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6757

## Description
Removed warning and js associated with it. 

## Screencast
https://streamable.com/xi56tp

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/interface.html
+++ b/interface.html
@@ -163,14 +163,12 @@
                 <label>When the form is submitted...</label>
               </div>
               <div>
-                <div class="checkbox checkbox-icon">
+                <div v-if="!settings.autobindProfileEditing" class="checkbox checkbox-icon">
                   <input type="checkbox" id="save_datasource" name="dataStore" v-model="settings.dataStore" value="dataSource">
                   <label for="save_datasource">
                     <span class="check"><i class="fa fa-check"></i></span> Add new submissions to a data source
                   </label>
                 </div>
-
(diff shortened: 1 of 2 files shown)


tonytlwu:
@squallstar See if you think this is appropriate.

js/libs/builder.js
```diff
@@ -86,6 +86,7 @@ function generateFormDefaults(data) {
     onSubmit: [],
     template: false,
     saveProgress: true,
+    autobindProfileEditing: false,
```
squallstar, line 89: why is this required?

interface.html
```diff
@@ -239,7 +237,7 @@ <h3>Submission</h3>
                 </div>
               </div>
             </div>
-            <div class="form-group">
+            <div  v-if="!settings.dataStore.length" class="form-group">
```
squallstar, line 240: have you confirmed dataStore is always an array and you don't need an explicit check to avoid this line failing?

MaksymShokin: datastore is either an empty array or an array with values

js/libs/builder.js
MaksymShokin: It was actually a mistake in not having it. We use this in some places

PR #322, opened 2020-08-06
[Feature] Allow users to add personalizations for the value of the field.

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6747

## Description
Allow users to add personalizations for the value of the field.

## Screenshots/screencasts
https://share.getcloudapp.com/kpubyvZE

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/components/checkbox.js
```diff
@@ -71,5 +71,10 @@ Fliplet.FormBuilder.field('checkbox', {
       this.value = [];
       this.updateValue(this.name, this.value);
     }
+  },
+  mounted: function () {
+    if (this.source !== 'defaultSource') {
+      this.setSourceValue({source: this.source, key: this.key});
```
tonytlwu, line 77:
```suggestion
      this.setSourceValue({ source: this.source, key: this.key });
```

Find other occurrences.

js/libs/core.js
```diff
@@ -67,6 +67,60 @@ Fliplet.FormBuilder = (function() {
...
+        var result;
+        var $vm = this;
+
+        switch (data.source) {
+          case 'profile':
+            if (!data.key) {
+              throw new Error('A key is required to fetch data from the user\'s profile');
+            }
+
+            result = Fliplet.Profile.get(data.key);
```
tonytlwu, line 80: User profile data should be loaded from `Fliplet.User.getCachedSession()`, not `Fliplet.Profile.get()`.

Once the session is retrieved, look for the profile key from the `dataSource` object in `sessions`. If it doesn't exist, look for it in the `saml2` object.

templates/configurations/checkbox.interface.hbs
```diff
@@ -1,8 +1,8 @@
+<div v-show="source === 'defaultSource'" class="form-group">
```
tonytlwu, line 1: **Default values** should come after **Description**. See https://github.com/Fliplet/fliplet-studio/issues/5813#issuecomment-668048849

templates/configurations/form.interface.hbs
```diff
@@ -84,6 +84,25 @@
       </div>
     {{/hasOptions}}
 
+    <div v-if="['flInput', 'flCheckbox', 'flRadio', 'flEmail', 'flNumber', 'flTelephone', 'flUrl', 'flTextarea', 'flWysiwyg', 'flSelect', 'flDate', 'flTime'].includes(_componentName)" class="form-group cleafix">
+      <div class="col-xs-12">
+        <label>Default value type</label>
+        <select class="form-control" :value="source" v-model="source">
+          <option value="defaultSource">Value</option>
+          <option value="profile">User's profile</option>
+          <option value="storage">App storage</option>
```
tonytlwu, line 93:
```suggestion
          <option value="appStorage">App storage</option>
```

templates/configurations/form.interface.hbs
```diff
@@ -84,6 +84,25 @@
...
+          <option value="query">Link query</option>
+        </select>
+        <div class="col-xs-12 text-right">
+          <span class="btn-link form-fields-customize">Learn more about these types <i class="text-muted fa fa-question-circle" ref="tooltip" data-viewport=".modal-body" data-toggle="tooltip" data-placement="left" title="Learn more about these types"></i></span>
```
tonytlwu, line 97: @sofiiakvasnevska @nishdandamudi Can you make sure CS provides support document for this so we can add a URL to link to?

@YaroslavOvdii We don't need a tooltip UI. Just a `<a>` linking to the support doc in a new tab will do.

templates/configurations/form.interface.hbs
```diff
@@ -84,6 +84,25 @@
...
+        </div>
+        <div v-show="source !== 'defaultSource'">
+          <label>Default key</label>
+          <input class="form-control" type="text" v-model.trim="key" placeholder="Specify property key" />
```
tonytlwu, line 101: Placeholder text should change according to the source type:

- `profile` Enter user data field/column name
- `appStorage` Enter storage key
- `query` Enter query name

js/libs/core.js
YaroslavOvdii: @tonytlwu 
Fliplet.Profile.get() - was taken from [here](https://github.com/Fliplet/fliplet-widget-form-builder/blob/f0f0fedf0792c810f7d77345068ae7d4c4a9e428/js/libs/form.js#L812)
Should we change it as well?

tonytlwu: @squallstar What are your thoughts on this, taking https://github.com/Fliplet/fliplet-studio/issues/5171 into consideration.

I imagine we don't want to change how field.set() works to avoid breaking changes, but I also don't want to start using `Fliplet.Profile.get()` for this new feature because I want to be able to load any user data from the session entries.

squallstar: @tonytlwu should we have fallbacks and load the first value we find for the key in these namespaces?

1. Session - dataSource passport
2. Session - saml2 passport
3. Session - any passport 
4. Profile

If the logic is clear enough in the future we can potentially port it to a new core JS API as we discussed in the ticket.

tonytlwu: @squallstar That works for me.

@YaroslavOvdii See above. And be sure to resolve conflicts before it's ready for review.

YaroslavOvdii: @squallstar @tonytlwu 
Do I get It right that we should look for the key as I show below?

image

squallstar: @YaroslavOvdii nope, you need to look into `session.entries`, see https://developers.fliplet.com/API/fliplet-session.html#read-details-about-connected-accounts

YaroslavOvdii: @squallstar I get it now, thank you.

YaroslavOvdii:
@tonytlwu Added requested changed, resolved conflicts.

js/libs/core.js
```diff
@@ -67,6 +67,76 @@ Fliplet.FormBuilder = (function() {
...
+          case 'query':
+            if (!data.key) {
+              throw new Error('A key is required to fetch data from the navigation query parameters');
+            }
+
+            result = Fliplet.Navigate.query[data.key];
+            break;
+          case 'appStorage':
+            if (!data.key) {
+              throw new Error('A key is required to fetch data from the storage');
```
tonytlwu, line 91:
```suggestion
              throw new Error('A key is required to fetch data from app storage');
```

js/libs/core.js
```diff
@@ -67,6 +67,76 @@ Fliplet.FormBuilder = (function() {
...
+          result = Promise.resolve(result);
+        }
+
+        return result.then(function (value) {
+          if (typeof value === 'undefined') {
+            value = '';
+          }
+
+          if (typeof value === 'object') {
+            if (value && value.entries) {
```
tonytlwu, line 110: This bit of the logic really only applies to when `result` comes from `Fliplet.User.getCachedSession()`, so I'd just add it as a `.then()` after the `Fliplet.User.getCachedSession()`.

js/libs/core.js
```diff
@@ -67,6 +67,76 @@ Fliplet.FormBuilder = (function() {
...
+              var entries = value.entries;
+
+              if (entries.dataSource) {
+                value = entries.dataSource.data[data.key]
+              } else if (entries.saml2) {
+                value = entries.saml2.data[data.key]
+              } else {
+                value = entries.flipletLogin.data[data.key]
```
tonytlwu, line 118: Where's the fallback to use `Fliplet.Profile.get()`?

YaroslavOvdii:
@tonytlwu  added requested changes, resolved conflicts.

js/libs/core.js
```diff
@@ -67,6 +67,77 @@ Fliplet.FormBuilder = (function() {
...
+            result = this.value;
+        }
+
+        if (!(result instanceof Promise)) {
+          result = Promise.resolve(result);
+        }
+
+        return result.then(function (value) {
+          if (typeof value === 'undefined') {
+            value = '';
```
tonytlwu, line 123: If the value is `undefined`, what's wrong with returning `undefined` as it is?

YaroslavOvdii: @tonytlwu for the most field I've worked with it is nothing, but it was taken from the original `set` function at this [line](https://github.com/Fliplet/fliplet-widget-form-builder/blob/a9ec4ffb962c5867676ce99046b53363ea467586/js/libs/form.js#L848) and I thought it is needed for some fields I haven't worked with.
Should I remove it?

tonytlwu: @squallstar What's your opinion? [You coded it](https://github.com/Fliplet/fliplet-widget-form-builder/blame/a9ec4ffb962c5867676ce99046b53363ea467586/js/libs/form.js#L848).

squallstar: @tonytlwu it may have been due to wanting to set the value to an object that has a common set of instance methods such as `indexOf`. The default value for an input field (at least a text one) is an empty string. If such field is accessing string instance methods then it would fail if we set that to undefined

tonytlwu: OK. Let's keep the enforced `""` then. @YaroslavOvdii Good job for finding this from `.set()` to keep our consistency.

js/components/wysiwyg.js
```diff
@@ -122,6 +122,10 @@ Fliplet.FormBuilder.field('wysiwyg', {
     }
   },
   mounted: function () {
+    if (this.source !== 'defaultSource') {
```
squallstar, line 125: couldn't this overly repeated logic be implemented in DRY fashion? I'm surprised we didn't manage to optimise this.

As a further insight, I think we would not have allowed this repetition on any other part of the system because we would have questioned it. cc @tonytlwu

templates/configurations/form.interface.hbs
```diff
@@ -84,6 +84,34 @@
       </div>
     {{/hasOptions}}
 
+    <div v-if="['flInput', 'flCheckbox', 'flRadio', 'flEmail', 'flNumber', 'flTelephone', 'flUrl', 'flTextarea', 'flWysiwyg', 'flSelect', 'flDate', 'flTime', 'flStarRating', 'flSignature', 'flImage', 'flFile'].includes(_componentName)" class="form-group cleafix">
```
squallstar, line 87: can this array be defined in code? it doesn't look easy to maintain now.

templates/configurations/form.interface.hbs
```diff
@@ -84,6 +84,34 @@
...
+      <div class="col-xs-12">
+        <label>Description</label>
+        <input class="form-control" type="text" v-model.trim="description" placeholder="Description text" />
+      </div>
+    </div>
+
+    <div v-if="['flInput', 'flCheckbox', 'flRadio', 'flEmail', 'flNumber', 'flTelephone', 'flUrl', 'flTextarea', 'flWysiwyg', 'flSelect', 'flDate', 'flTime'].includes(_componentName)" class="form-group cleafix">
```
squallstar, line 94: same as above

YaroslavOvdii:
@squallstar `defaultSource` it's a condition when user decide to set default value as it was before this feature. We called it like that because it is also a initial value for the `source` variable.

js/components/wysiwyg.js
YaroslavOvdii: @squallstar We could move this logic to the separate function in the core.js file and in this way it would be common to all components and we can call it in components that we are needed.
Will it be good to do that way?

squallstar: @YaroslavOvdii what if the mounted function is defined by the system and the component's is called afterwards? This has been done for many other occasions in the form:

```
var mounted = component.mounted;
component.mounted = function () {
  // do something
  mounted.call(this);
}
```

squallstar:
@YaroslavOvdii what is the default source is simply no source? e.g. empty string.

Alternatively, the default value should be `default`, not `defaultSource`. There is no need to repeat the key name. @tonytlwu do you agree?

tonytlwu:
@squallstar I agree. `default` should be sufficient, and even better if we can just use `!source` for empty string or undefined value.

js/components/wysiwyg.js
YaroslavOvdii: @squallstar at the moment we didn't use defined by the system mounted function because we were needed that our new feature work right after the component was loaded, but system mount function was added only after the component appears on the screen and it doesn't call mount again to use it. That is why in all components that must have this feature we have added mount function.

squallstar: @YaroslavOvdii I don't understand your last comment, but more importantly please just look at code that we wrote in the past if you need to see how we have created a default mounted method in the basic library https://github.com/Fliplet/fliplet-widget-form-builder/blob/master/js/libs/core.js#L272

PR #327, opened 2020-08-19
[Fix] Corrected autofill behavior

MaksymShokin:
@sofiiakvasnevska

## Issue
Fliplet/fliplet-studio#5813

## Description
Added force autofill if "autofill always" selected after date/time update.

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/js/components/date.js
+++ b/js/components/date.js
@@ -40,7 +40,9 @@ Fliplet.FormBuilder.field('date', {
   },
   methods: {
     updateValue: function(value) {
-      if (value) {
+      if (this.autofill === 'always') {
+        this.value = moment().format('YYYY-MM-DD');
+      } else if (value) {
         this.value = value;
       }
 
--- a/js/components/time.js
+++ b/js/components/time.js
@@ -33,12 +33,24 @@ Fliplet.FormBuilder.field('time', {
   },
   methods: {
     updateValue: function(value) {
-      if (value) {
+      if (this.autofill === 'always') {
+        this.value = this.getCurrentTime();
+      } else if (value) {
         this.value = value;
       }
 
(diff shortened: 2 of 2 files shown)


js/components/date.js
```diff
@@ -40,7 +40,9 @@ Fliplet.FormBuilder.field('date', {
   },
   methods: {
     updateValue: function(value) {
-      if (value) {
+      if (this.autofill === 'always') {
```
tonytlwu, line 43: It seems to me that this suggests if user tries to change the value of a date field, given `autofill === 'always'`, they won't be able to change the value because it'll always be forced to `moment().format('YYYY-MM-DD')`. If so, this is incorrect.

> **Always autofill with current date/time** When this is checked, set the field with today's date after the form is loaded, regardless if the field has a value loaded.

It meets the requirement above, but it shouldn't lock the value down. Users should still be able to change it afterwards as per the following requirement.

> **Use the date when form is loaded** This would be the current behaviour. The field is loaded with Date.now() when the form is loaded. _Users may change this value unless **Read-only** is selected._ This is the default checked option.

tonytlwu (changes requested):

js/components/date.js
MaksymShokin: @tonytlwu 
In this case, this comment mentioned by Inna is not an issue https://github.com/Fliplet/fliplet-studio/issues/5813#issuecomment-675547497 ?

tonytlwu: @MaksymShokin It is an issue. See https://github.com/Fliplet/fliplet-studio/issues/5813#issuecomment-680771881 for clarification.

PR #330, opened 2020-09-17
[Feature] Integrated data source provider into the form builder

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-10

## Description
Integrated data source provider into the form builder

## Screenshots/screencasts
https://share.getcloudapp.com/7KuLKQLx

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/js/libs/builder.js
+++ b/js/libs/builder.js
@@ -632,14 +632,11 @@ var app = new Vue({
     initDataSourceProvider: function() {
       var $vm = this;
       var dataSourceData = {
-        dataSourceTitle: 'Form builder data source',
+        dataSourceTitle: 'Form data source',
         dataSourceId: $vm.settings.dataSourceId,
         appId: Fliplet.Env.get('appId'),
         default: {
-          name: 'Form builder data for ' + Fliplet.Env.get('appName'),
-          definition: {
-            'guid': 'Column 1'
-          },
+          name: 'Form data for ' + $vm.settings.displayName,
           entries: [],
           columns: []
         },
--- a/js/libs/builder.js
+++ b/js/libs/builder.js
@@ -144,7 +144,8 @@ var app = new Vue({
       showDataSourceSettings: !!formSettings.dataSourceId,
       organizationName: '',
       isPreviewing: formSettings.previewingTemplate !== '',
-      editor: undefined
+      editor: undefined,
+      accessRulesTypes: ['insert']
     };
   },
   methods: {
@@ -246,54 +247,6 @@ var app = new Vue({
       if (this.isAddingFields) {
         Fliplet.Studio.emit('widget-mode', 'wide');
       }
-    },
-    createDataSource: function() {
-      var $vm = this;
-
-      Fliplet.Modal.prompt({
-        title: 'Please enter a data source name'
-      }).then(function(result) {
-        if (result === null) {
-          $vm.settings.dataSourceId = '';
-          return;
-        }
-
-        var dataSourceName = result.trim();
-
-        if (!dataSourceName) {
-          Fliplet.Modal.alert({
-            message: 'You must enter a data source name'
-          }).then(function() {
-            $vm.settings.dataSourceId = '';
-            $vm.createDataSource();
-            return;
-          });
-        }
-
-        Fliplet.DataSources.create({
-          name: dataSourceName,
-          organizationId: Fliplet.Env.get('organizationId')
-        }).then(function(ds) {
-          $vm.dataSources.push(ds);
-          $vm.settings.dataSourceId = ds.id;
-          $vm.showDataSourceSettings = true;
-        });
-      });
-    },
-    manageDataSource: function(dataSourceId) {
-      Fliplet.Studio.emit('overlay', {
-        name: 'widget',
-        options: {
-          size: 'large',
-          package: 'com.fliplet.data-sources',
-          title: 'Edit Data Sources',
-          classes: 'data-source-overlay',
-          data: {
-            context: 'overlay',
-            dataSourceId: dataSourceId
-          }
-        }
-      });
     },
     save: function(initial) {
       var $vm = this;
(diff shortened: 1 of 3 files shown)


js/libs/builder.js
```diff
@@ -686,6 +629,47 @@ var app = new Vue({
 
       return string;
     },
+    initDataSourceProvider: function() {
+      var $vm = this;
+      var dataSourceData = {
+        dataSourceTitle: 'Form builder data source',
```
tonytlwu, line 635:
```suggestion
        dataSourceTitle: 'Form data source',
```

js/libs/builder.js
```diff
@@ -686,6 +629,47 @@ var app = new Vue({
...
+        dataSourceId: $vm.settings.dataSourceId,
+        appId: Fliplet.Env.get('appId'),
+        default: {
+          name: 'Form builder data for ' + Fliplet.Env.get('appName'),
+          definition: {
```
tonytlwu, line 640: This doesn't need a `guid`.

js/libs/builder.js
```diff
@@ -686,6 +629,47 @@ var app = new Vue({
...
       return string;
     },
+    initDataSourceProvider: function() {
+      var $vm = this;
+      var dataSourceData = {
+        dataSourceTitle: 'Form builder data source',
+        dataSourceId: $vm.settings.dataSourceId,
+        appId: Fliplet.Env.get('appId'),
+        default: {
+          name: 'Form builder data for ' + Fliplet.Env.get('appName'),
```
tonytlwu, line 639:
```suggestion
          name: 'Form data for ' + formName,
```

js/libs/builder.js
```diff
@@ -686,6 +629,47 @@ var app = new Vue({
...
+          definition: {
+            'guid': 'Column 1'
+          },
+          entries: [],
+          columns: []
+        },
+        accessRules: [
```
tonytlwu, line 646: Incorrect access rules. See https://docs.google.com/presentation/d/1vvEvfGm_XGjKE3vp1duMy8DuqrgQGD3Dvlf4M-xNQQU/edit?ts=5f341a02#slide=id.g8d5e20b4f1_0_0

@upplabs-alex-levchenko @sofiiakvasnevska Please follow the specs provided. I shouldn't need to open the document to check this myself.

tonytlwu (changes requested):
@squallstar Could you review this too as you were involved in the initial build of this component?

js/libs/builder.js
YaroslavOvdii: @tonytlwu it said in the spec that form builder rules are: `Write, Update - only if the update is enabled` and that is exactly how rules are set. Is there something more that should be added to these rules?
image

js/libs/builder.js
YaroslavOvdii: I'll remove it then.

js/libs/builder.js
tonytlwu: @YaroslavOvdii Yes, that is correct. Sorry this was my bad, I saw `"allow": "all"` and it looked like we're allowing them to do everything, but I was mistaken. This is fine.

tonytlwu:
@YaroslavOvdii Looks good to me now. I'd like @squallstar to review and give it the OK too.

js/libs/builder.js
```diff
@@ -144,7 +144,8 @@ var app = new Vue({
       showDataSourceSettings: !!formSettings.dataSourceId,
       organizationName: '',
       isPreviewing: formSettings.previewingTemplate !== '',
-      editor: undefined
+      editor: undefined,
+      accessRulesTypes: ['insert']
```
squallstar, line 148: what's this for and why doesn't it have "update" too?

js/libs/builder.js
```diff
@@ -686,6 +629,44 @@ var app = new Vue({
...
       return string;
     },
+    initDataSourceProvider: function() {
+      var $vm = this;
+      var dataSourceData = {
+        dataSourceTitle: 'Form data source',
+        dataSourceId: $vm.settings.dataSourceId,
+        appId: Fliplet.Env.get('appId'),
+        default: {
+          name: 'Form data for ' + $vm.settings.displayName,
```
squallstar, line 639: @tonytlwu has this name been agreed?

js/libs/builder.js
```diff
@@ -832,6 +809,14 @@ var app = new Vue({
     'settings.dataStore': function(value) {
       this.showExtraAdd = value.indexOf('dataSource') > -1;
       this.showExtraEdit = value.indexOf('editDataSource') > -1;
```
squallstar, line 811: what does showExtraEdit mean? I also don't see it defined in the data object

squallstar (changes requested):
Reviewed

js/libs/builder.js
tonytlwu: Yes. I brought it up in a previous comment. It's been updated based on my suggestions.

js/libs/builder.js
YaroslavOvdii: @squallstar If I understand it right it is showing an extra setting when we click on ` Update existing entries in a data source` checkbox.
image
Same thing for `showExtraAdd `.

js/libs/builder.js
YaroslavOvdii: @squallstar I've added the `accessRulesTypes` variable so we could change its value depending on what setting is set.
In this [file](https://docs.google.com/presentation/d/1vvEvfGm_XGjKE3vp1duMy8DuqrgQGD3Dvlf4M-xNQQU/edit#slide=id.g8d5e20b4f1_0_0) said that we should set the data source rule `update` only if this `Update existing entries in a data source` checkbox is selected. That is why when we click on it we change our variable in what state it should be. You may see it at [this](https://github.com/Fliplet/fliplet-widget-form-builder/pull/330/files#diff-892056c1c88dfac81f33ecf67625a5ecR812) line.

squallstar (approved):

squallstar:
Approved.

tonytlwu (approved):

Risk: Critical
Affected: Form components configured in Studio
Verify: Data configurations are correctly saved and loaded

PR #332, opened 2020-10-05
[Fix] OD-122 Datepicker will show correct date when we use field.set()

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-122

## Description
Update value for the date every time

## Screenshots/screencasts
https://share.getcloudapp.com/p9uwj06R

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -856,7 +856,7 @@ Fliplet.Widget.instance('form-builder', function(data) {
 
                   // To set correct date in the date picker
                   // We want to trigger change on date field every time.
-                  if (hasChanged || !field._type === 'flDate') {
+                  if (hasChanged || field._type === 'flDate') {
                     $form.triggerChange(field.name, field.value);
                   }
                 });
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -854,7 +854,9 @@ Fliplet.Widget.instance('form-builder', function(data) {
                   field.value = value;
                   debouncedUpdate();
 
-                  if (hasChanged) {
+                  // To set correct date in the date picker
+                  // We want to trigger change on date field every time.
+                  if (hasChanged || !field._type === 'flDate') {
                     $form.triggerChange(field.name, field.value);
                   }
                 });


js/libs/form.js
```diff
@@ -854,7 +854,9 @@ Fliplet.Widget.instance('form-builder', function(data) {
                   field.value = value;
                   debouncedUpdate();
 
-                  if (hasChanged) {
+                  // To set correct date in the date picker
+                  // We want to trigger change on date field every time.
+                  if (hasChanged || !field._type === 'flDate') {
```
squallstar, line 859: You sure about that exclamation mark? I've never seen such block condition statement.

squallstar (changes requested):

js/libs/form.js
YaroslavOvdii: @tonytlwu As I understand the work of the date picker in the date component correct, we should trigger a change every time we do a set() to tell the date picker set a correct date.
Should I remove this comment?

tonytlwu: @YaroslavOvdii I don't think @squallstar's question has anything to do with requirements. Your syntax looks wrong and possibly untested.

squallstar: @YaroslavOvdii have you tested this? @tonytlwu yes that's what I was referring to, I thought my comment was explicit enough but it probably wasn't.

@YaroslavOvdii to confirm, `!field._type` would produce `false`, and `false` is never going to equal `flDate`.

YaroslavOvdii: @squallstar Ou I get it now, sorry for the misunderstanding.
Yes, I have tested it but it would be better to change it to `field._type === date`

squallstar: @YaroslavOvdii you sure you tested it and it works? From this fiddle it doesn't look like: https://jsfiddle.net/qkdto6L7/

YaroslavOvdii: @squallstar Yes, you are right about that, it worked for me because I didn't test it correctly. I've tested it with connection with the LFD in the chain LFD -> edit item -> look a form builder. In that case `field.set()` will be working correctly.
If we use `field._type === 'flDate'` it will be working correctly for all other cases. Sorry for the long answer.

squallstar (approved):

PR #333, opened 2020-10-09
ID-328 [Feat] Support for sending image names via base64

squallstar:
requires https://github.com/Fliplet/fliplet-api/pull/4332 to be deployed before this is released.

### DIFF
--- a/js/components/image.js
+++ b/js/components/image.js
@@ -200,11 +200,12 @@ Fliplet.FormBuilder.field('image', {
 
           var scaledImage = loadImage.scale(img, options);
           var imgBase64Url = scaledImage.toDataURL(mimeType, $vm.jpegQuality);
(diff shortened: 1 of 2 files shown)


js/components/image.js
```diff
@@ -200,11 +200,12 @@ Fliplet.FormBuilder.field('image', {
 
           var scaledImage = loadImage.scale(img, options);
           var imgBase64Url = scaledImage.toDataURL(mimeType, $vm.jpegQuality);
+          var flipletBase64Url = imgBase64Url + ';filename:' + file.name;
```
tonytlwu, line 203: Does this assume there will always be a file name and is that a valid assumption?

Or does this file set a default `file.name` somewhere?

squallstar: @tonytlwu https://developer.mozilla.org/en-US/docs/Web/API/File/name

tonytlwu (approved):

PR #343, opened 2020-11-04
ID-379 [Feat] Adapt field data so checkbox and non-checkbox fields can support more data types

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-379

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -133,14 +133,22 @@ Fliplet.Widget.instance('form-builder', function(data) {
 
           var showCurrentDateTime = field.autofill === 'always';
 
-          if (field._type === 'flCheckbox') {
-            if (!Array.isArray(fieldData)) {
-              fieldData = _.compact([fieldData]);
-            }
-          } else {
-            if (Array.isArray(fieldData)) {
-              fieldData = fieldData[0];
-            }
+          switch (field._type) {
+            case 'flCheckbox':
+              if (!Array.isArray(fieldData)) {
+                fieldData = _.compact([fieldData]);
+              }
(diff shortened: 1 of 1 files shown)


squallstar (dismissed):
- Risk: high, do you agree?

js/libs/form.js
```diff
@@ -133,6 +133,16 @@ Fliplet.Widget.instance('form-builder', function(data) {
 
           var showCurrentDateTime = field.autofill === 'always';
 
+          if (field._type === 'flCheckbox') {
```
squallstar, line 136: why don't you move this to the switch case?

js/libs/form.js
```diff
@@ -133,6 +133,16 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
+            if (!Array.isArray(fieldData)) {
+              fieldData = _.compact([fieldData]);
+            }
+          } else {
+            if (Array.isArray(fieldData)) {
```
squallstar, line 141: Wouldn't this break checkboxes and files?

tonytlwu: It shouldn't break checkboxes because checkbox fields won't come to this branch of the code, but files (and images) might need to be excluded too.

js/libs/form.js
tonytlwu: The `fieldData` typecasting needs to happen before we go into the switch cases.

tonytlwu:
@squallstar Yes, this is a high risk as it could affect all field types.

squallstar (approved):
👏🏻

PR #350, opened 2020-11-23
OD-200 [Feature] Showing a modal on form builder configuration change if there are missing security rules.

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-200

## Description
Showing a modal on form builder configuration change if there are missing security rules.

## Screenshots/screencasts
https://share.getcloudapp.com/Qwubxo2W

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/js/libs/builder.js
+++ b/js/libs/builder.js
@@ -590,6 +590,7 @@ new Vue({
     useTemplate: function(templateId) {
       Fliplet.Studio.emit('widget-save-label-reset');
       Fliplet.Widget.toggleSaveButton(true);
+
       var $vm = this;
 
       this.updateFormSettings(templateId, false);
@@ -614,11 +615,13 @@ new Vue({
 
       if (this.chooseTemplate && preview) {
         this.settings.previewingTemplate = templateId;
+
         return;
       }
 
       if (this.isPreviewing) {
         this.settings.previewingTemplate = '';
+
         return;
       }
     },
@@ -628,6 +631,7 @@ new Vue({
 
       if (more) {
         $vm.readMore.push(templateId);
+
         return;
       }
 
@@ -767,7 +771,7 @@ new Vue({
             if ($vm.settings.resultHtml) {
               var updatedHtml = $vm.convertVueEventAttributes($vm.settings.resultHtml);
 
-              editor.setContent(updatedHtml, {format: 'raw'});
+              editor.setContent(updatedHtml, { format: 'raw' });
             }
           });
 
@@ -797,6 +801,7 @@ new Vue({
           }
         });
       });
+
       return $html.html();
     },
     loadTemplates: function() {
@@ -994,6 +999,7 @@ new Vue({
     window.emailTemplateEditProvider = null;
     window.generateEmailProvider = null;
     window.linkProvider = null;
+
     var $vm = this;
 
     $vm.settings.name = $vm.settings.name || 'Untitled form';
--- a/js/libs/builder.js
+++ b/js/libs/builder.js
@@ -148,7 +148,13 @@ new Vue({
       organizationName: '',
       isPreviewing: formSettings.previewingTemplate !== '',
       editor: undefined,
-      accessRulesTypes: ['insert']
+      accessRules: [
+        {
+          allow: 'all',
+          enabled: true,
+          type: ['select']
+        }
+      ]
     };
   },
   methods: {
@@ -648,13 +654,7 @@ new Vue({
           entries: [],
           columns: []
         },
-        accessRules: [
-          {
-            allow: 'all',
-            enabled: true,
-            type: $vm.accessRulesTypes
-          }
-        ]
+        accessRules: $vm.accessRules
       };
 
       window.dataSourceProvider =  Fliplet.Widget.open('com.fliplet.data-source-provider', {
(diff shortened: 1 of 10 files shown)


js/libs/builder.js
```diff
@@ -708,6 +708,15 @@ new Vue({
         $vm.triggerSave();
       });
     },
+    toggleAccessType: function(type, isTypeActive) {
+      var typeIndex = this.accessRules[0].type.indexOf(type);
```
tonytlwu, line 712: Why are we only looking for the first one rather than searching the entire list of access rules?

And if `this.accessRules` is empty, `this.accessRules[0].type` will result in an error.

tonytlwu (changes requested):

js/libs/builder.js
YaroslavOvdii: @tonytlwu `this.accessRules` is a default variable set by us and it's always present and always has only one element in it. Will it be changed in the future?

tonytlwu: I don't know if it will change in the future, but a search within the entire list will not change the output behaviour and shouldn't add much more complexity. Let'd do that.

YaroslavOvdii: @tonytlwu While I was implementing this solution question appeared.
As you may see in the next lines we are using the same way to add or remove a rule type. But how we should determine to what access rules we should add or remove the new type? Should we do it for all of them in for each loop?

tonytlwu: @YaroslavOvdii See Slide 11

* If any of the required security rules is missing (based on active features), show the security rule warning.
* The security rules are verified by the **type** of each rule, not by **user condition**.
* When a user deactivates a feature requires a security rule:
    * Security rules are not assessed or updated or removed (by design)
        * Long term solution is to add security check before launch


And Slide 12

* Check the data source security rules to make sure the security rule warning is shown if any of the required rules (based on the active features) is missing.

---

We should always check directly if the data source has the right access rules based on the rule types. And rules don't need to be removed or updated of the activated features are changed. We just need to make sure we don't add multiple of the same type.

YaroslavOvdii:
@tonytlwu added suggested changes.
New [screencast](https://share.getcloudapp.com/RBu9wz9B)

tonytlwu:
@YaroslavOvdii Can you make sure the branch is up to date with the base branch first?

YaroslavOvdii:
@tonytlwu Updated

tonytlwu:
@YaroslavOvdii It seems that the structure you've gone for is outlined as below:

1. Form component (or any component that uses the data source provider) maintains a list of security rules and runs operations that updates what the list of security rules should be.
1. Form component sends the list of security rules to the Data Source Provider (DSP)
1. DSP takes the instruction and updates the data source with security rules

If so, I don't think this is right. We should aim for a structure outlined as below:

1. Form component sends instruction to DSP specifying a specific rule (with specific user condition) is required
1. DSP takes the request, assesses the data source to determine if a new rule needs to be added
1. If it doesn't need to be added, nothing happens
1. If it needs to be added, it tells the user that it's needed and the user decides whether to update the list of data source security rules accordingly

squallstar:
@tonytlwu I agree with your recommended implementation.

js/libs/builder.js
```diff
@@ -712,6 +712,17 @@ new Vue({
         $vm.triggerSave();
       });
     },
+    toggleAccessType: function(type, isTypeActive) {
+      this.accessRules.forEach(function(rule, index, accessRules) {
+        var typeIndex = rule.type.indexOf(type);
+
+        if (isTypeActive && typeIndex === -1) {
+          accessRules[index].type.push(type);
```
tonytlwu, line 720: You shouldn't be modifying the `type` array for each security rule. Instead, add a new rule to `accessRules`.

YaroslavOvdii:
@tonytlwu Actually it is working as you've described here:

> 1.Form component sends instruction to DSP specifying a specific rule (with specific user condition) is required
> 2. DSP takes the request, assesses the data source to determine if a new rule needs to be added
> 3. If it doesn't need to be added, nothing happens
> 4. If it needs to be added, it tells the user that it's needed and the user decides whether to update the list of data source security rules accordingly

Form builder and LFD sends a different scope of types that are needed for the component (e.g Read, Write, Update, etc)
And in DSP we are checking if we need to update rules and take actions after that check.
It is done in this [PR](https://github.com/Fliplet/fliplet-widget-data-source-provider/pull/25)

js/libs/builder.js
YaroslavOvdii: @tonytlwu Ok, I'll change it.

tonytlwu:
@YaroslavOvdii If that is the case, I don't think it needs to send `enabled: true` to DSP.

YaroslavOvdii:
@tonytlwu We also have `allow: all`, as a common option in all components. Should we move it to the provider as well or better leave it as it is?

tonytlwu:
@YaroslavOvdii I would leave it there. The `allow` property could be different from feature to feature, so let's keep that.

YaroslavOvdii:
@tonytlwu Added requested changes

tonytlwu (approved):

Risk: High
Affected: Studio users configuring form data sources
Verify: Users can configure form settings

Squash and merge: Update security rules

PR #351, opened 2020-11-23
[Feat] Adds support for query in Fliplet.FormBuilder.getAll()

tonytlwu:
Adds similar support to https://github.com/Fliplet/fliplet-api/pull/4384 for using `Fliplet.FormBuilder.getAll(name)` with a query.

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -1016,11 +1016,11 @@ Fliplet.FormBuilder.get = function (name) {
 Fliplet.FormBuilder.getAll = function (name) {
   return Promise.all(formBuilderInstances).then(function(forms) {
     if (typeof name === 'undefined') {
-      return forms
+      return forms;
     }
 
-    return forms.filter(function(vueForm) {
-      return vueForm.name === name;
+    return forms.filter(function(form) {
+      return form.name === name;
     });
   });
 };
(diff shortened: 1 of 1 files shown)


js/libs/form.js
```diff
@@ -1013,6 +1013,14 @@ Fliplet.FormBuilder.get = function (name) {
   });
 };
 
-Fliplet.FormBuilder.getAll = function () {
-  return Promise.all(formBuilderInstances);
+Fliplet.FormBuilder.getAll = function (name) {
+  return Promise.all(formBuilderInstances).then(function(forms) {
+    if (typeof name === 'undefined') {
+      return forms
```
squallstar, line 1019: missing semicolon

js/libs/form.js
```diff
@@ -1013,6 +1013,14 @@ Fliplet.FormBuilder.get = function (name) {
...
+    }
+
+    return forms.filter(function(vueForm) {
```
squallstar, line 1022: I don't think it's necessary to have the `vue` prefix here

tonytlwu: I don't think so either. I was following the previous pattern. I'll remove the prefix.

tonytlwu:
@squallstar Ready for review

squallstar (approved):

PR #354, opened 2020-12-04
OD-282 [Fix] Security rules are missing after clicking the "I'll do it later" 

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-282 https://weboo.atlassian.net/jira/software/projects/OD/boards/9?selectedIssue=OD-282
## Description
Missing security rules notification is not showing after safe/close lfd component
## Screenshots/screencasts
https://screencast-o-matic.com/watch/cYlfXfNUX6
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko


js/libs/builder.js
```diff
@@ -1035,6 +1035,9 @@ new Vue({
       }
     }, savedLinkData);
 
+    this.toggleAccessType('insert', this.showExtraAdd);
```
tonytlwu, line 1038:
```js
      this.toggleAccessType('insert', this.showExtraAdd);
      this.toggleAccessType('update', this.showExtraEdit);
```

This is called twice in this file. I think they should be collated into a `this.updateAccessTypes()` function. This way when we have new access rules to consider, there's only one place to update.

tonytlwu (changes requested):

js/libs/builder.js
AndrRyaz: @tonytlwu, i create the updateAccessTypes function and moved toggling of rules, there

tonytlwu (approved):

Risk: Medium
Affected: Studio users configuring Form component
Verify: Data source security rules are always correct

Squash and merge: Update data source security rules

PR #355, opened 2020-12-29
OD-177 [Feature] Allow users to add personalization's for the value of the field

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-177

## Description
I've created a new PR from [this one](https://github.com/Fliplet/fliplet-widget-form-builder/pull/322) to avoid any conflicts in a code.
So we have added personalization for the default value of the fields based on the `.set()` function from the FormBuilder API.
All previous requests are added to this PR.
There will be several videos with more detailed explanations of how it works.

## Screenshots/screencasts
https://share.getcloudapp.com/wbuKLre9

## Backward compatibility

## Reviewers
@AndrRyaz @ivandevupp


js/libs/core.js
```diff
@@ -75,6 +75,76 @@ Fliplet.FormBuilder = (function() {
         };
       }
 
+      component.methods.setSourceValue = function(data) {
+        var result;
+        var $vm = this;
+
+        switch (data.source) {
```
squallstar, line 82: Has this been copied from somewhere else? I'm asking because if that's the case we may be able to put this logic into a shared file, e.g. `fliplet-core`

YaroslavOvdii: @squallstar Yes, it was copied from this [function ](https://github.com/Fliplet/fliplet-widget-form-builder/blob/master/js/libs/form.js#L840). But we changed it a little for our case (striped case function and literal from it.)

js/libs/core.js
```diff
@@ -161,6 +237,14 @@ Fliplet.FormBuilder = (function() {
         canHide: {
           type: Boolean,
           default: true
+        },
+        source: {
```
squallstar, line 241: We should be more specific in what these represent, e.g. `defaultValueSource` and `defaultValueKey`. @tonytlwu do you agree?

js/libs/core.js
```diff
@@ -75,6 +75,76 @@ Fliplet.FormBuilder = (function() {
         };
       }
 
+      component.methods.setSourceValue = function(data) {
```
squallstar, line 78: You are setting the value given the source and key, not setting the value of the source so this method should be renamed, e.g. `setValueFromDefaultSettings`

squallstar, line 78: Please add a JSDoc for this method

js/libs/core.js
```diff
@@ -213,6 +301,21 @@ Fliplet.FormBuilder = (function() {
         type: Array
       };
 
+      component.props._componentName = {
```
squallstar, line 304: What is the reason for these three variables to be defined as props? Can't they be just plain variables defined at the top of core.js?

templates/configurations/form.interface.hbs
```diff
@@ -84,6 +84,37 @@
...
+        </select>
+        <div class="col-xs-12 text-right">
+          <a href="#!" target="_blank">Learn more about these types</a>
+        </div>
+        <div v-show="source !== 'default'">
+          <label>Default key</label>
+          <div v-show="!key" class="text-danger">
+            This field is required*
+          </div>
+          <input v-bind:class="{ 'border-danger': !key }" v-show="source === 'profile'"class="form-control" type="text" v-model.trim="key" placeholder="Enter user data field/column name" />
```
squallstar, line 111: any specific reason for using `v-show` instead of `v-if`?

YaroslavOvdii: @squallstar No, should we use `v-if` instead?

js/libs/core.js
squallstar: Can the two functions be aggregated into a single one which we can then move as required later? We should be always working towards keeping code DRY.

templates/configurations/form.interface.hbs
squallstar: https://v1.vuejs.org/guide/conditional.html

> The difference is that an element with v-show will always be rendered and remain in the DOM; v-show simply toggles the display CSS property of the element.

As you can see, `v-if` is better because the node only gets rendered as required. On the other hand, use `v-show` when removing and adding the node can result in problems (e.g. because of added plugins on top of the rendering).

js/libs/core.js
YaroslavOvdii: @squallstar Yes, it is possible. Should I change it so it would work in both cases?

squallstar: Yes please.

js/libs/core.js
tonytlwu: `defaultValueSource` would be better indeed.

js/libs/core.js
YaroslavOvdii: @squallstar While I've tried to aggregated functions into one I've stuck with a question.
To get data from the Profile in the `.set()` function we use this [code](https://github.com/Fliplet/fliplet-widget-form-builder/blob/master/js/libs/form.js#L861) 
But in our new feature, we will use this [code](https://github.com/Fliplet/fliplet-widget-form-builder/pull/355/files#diff-962c976e04650f0e20b7b5a03282b499b82960438ed596001139d959084b8e2fR88)
In aggregated function what should be left?
And if we take a new example won't it make trouble for the current users?

squallstar: @YaroslavOvdii I think you should use the newest function's logic but with the following change: `Fliplet.Profile.get(data.key);` should always be used as a fallback when the data is not found in the target passport, e.g.

```
value = value from dataSource or saml2 or flipletLogin session
if value is undefined, value = Fliplet.Profile.get(data.key);
return value
```

Does it make sense?

@tonytlwu do you agree with this?

tonytlwu: @squallstar What you've outlined here makes sense. The LFD was updated with a similar logic.

See https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/464/files#diff-1e20a0ee0fa5c5689cd8541a5892e5e23248b3bf2ab918043e525c443856f18c

YaroslavOvdii: @squallstar I'm not sure where should I put the aggregated function so I put it in this comment.
```
/**
 * This function is responsible for getting the value from the source that was specified by the user
 * 
 * @param {Object} data - data that was passed by a user into the set function
 * 
 * @return {Promise} - returns promise with data that was returned from the source
 */
function getSourceData(data) {
  var result;
  var data = options.data;

  if (typeof data === 'function') {
    data = { source: 'function', key: data };
  }

  if (typeof data !== 'object') {
    data = { source: 'literal', key: data };
  }

  switch (data.source) {
    case 'profile':
      if (!data.key) {
        throw new Error('A key is required to fetch data from the user\'s profile');
      }

      result = Fliplet.User.getCachedSession()
        .then(function(session) {
          if (session && session.entries) {
            if (session.entries.dataSource) {
              return session.entries.dataSource.data[data.key];
            }

            if (session.entries.saml2) {
              return session.entries.saml2.data[data.key];
            }

            if (session.entries.flipletLogin) {
              return session.entries.flipletLogin.data[data.key];
            }
          }

          return Fliplet.Profile.get(data.key);
        });
      
        break;
    case 'query':
      if (!data.key) {
        throw new Error('A key is required to fetch data from the user\'s profile');
      }
      
      result = Fliplet.Navigate.query[data.key];
      
      break;
    case 'storage':
    case 'appstorage':
      if (!data.key) {
        throw new Error('A key is required to fetch data from the user\'s profile');
      }

      var storage = data.source === 'storage'
        ? Fliplet.Storage
        : Fliplet.App.Storage;
      result = storage.get(data.key);
      
      break;
    case 'function':
      result = data.key();
      
      break;
    case 'literal':
      result = data.key;
      
      break;
    default:
      result = data;

      break;
  }

  if (!(result instanceof Promise)) {
    result = Promise.resolve(result);
  }

  return result;
}
```
How it should work:
When I've analyzed the `set` and my functions I understand that the above part is the same for both of them so I've separated it and when we put it into the `fliplet-core` we should be able to work with the returned value as we had previously but without duplicating the code.

YaroslavOvdii:
@squallstar added requested changes.

js/libs/core.js
```diff
@@ -75,7 +75,17 @@ Fliplet.FormBuilder = (function() {
...
-      component.methods.setSourceValue = function(data) {
+      /**
+       * This function is responsible personalizating a default value when we need to get it from user personalized data
+       * For example: set default value to the user email
+       *
+       * @param {Object} data - an object with source and key properties
+       *  Where source - is defining the value  source type
+       *  And key - tell us what key in the source we should take as a value
+       *
+       * @return {void} - this function is not returns anything but updated value directlly with the help of the updateValue function.
```
squallstar, line 86: comment: this method does not return anything but updates the value directly on the target field

YaroslavOvdii: @squallstar updated comment

squallstar (approved):
@sofiiakvasnevska this is ready to be deployed on **staging** for QA.

- **Risk**: Critical
- **Affected**: All forms
- **Verify**: Verify forms can be seen and can be submitted

PR #358, opened 2021-01-12
OD-127 [Fix] Parse submit errors correctly

ivandevupp:
@sofiiakvasnevska
## Issue
OD-127 https://weboo.atlassian.net/browse/OD-127
## Description
Submit errors (from backend) are parsed correctly (not as cryptic and code-like).
## Screenshots/screencasts
https://monosnap.com/file/zexu8Dx5jcKnBe9XUpnnLcRuDsE175
## Backward compatibility
## Reviewers 
@upplabs-alex-levchenko @AndrRyaz


js/libs/form.js
```diff
@@ -639,8 +639,7 @@ Fliplet.Widget.instance('form-builder', function(data) {
 
           $vm.loadEntryForUpdate();
         }, function(err) {
-          console.error(err);
```
squallstar, line 642: why did you remove the `console.error`?

ivandevupp: I took a look at other catch structures in widget, and found out that we don't use console there except one place in `image.js` which was added 4 years ago. Also I have a habit to clear consoles before delivering tasks.
If I it is useful for testing/Sentry I'll bring it back now.

squallstar: Please bring it back, this is useful when clients and their developers are debugging via Fliplet Studio

ivandevupp: Done.

squallstar (approved):
@sofiiakvasnevska this is ready to be deployed on **staging** for QA.

- **Risk**: Low
- **Affected**: Forms when displaying errors
- **Verify**: Verify errors are correctly displayed

PR #363, opened 2021-02-10
OD-237 [Feature] Users will be able to set fields as read-only

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-237

## Description
Added a read-only feature for users to use.
Some fields like Image upload, File upload, Rich text has different from the others field view in the read-only mode.

## Screenshots/screencasts
https://share.getcloudapp.com/ApuRklPj - part 1
https://share.getcloudapp.com/WnuBJpXx - part 2
https://share.getcloudapp.com/BluY52v1 - part 3 (about developer options)

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

## Notes
Also a suggestion:
We could change these fields view as well:
URL - show an `a` tag instead of the simple text;
Email - made it interactive and use send email instead of the simple text;
Telephone -  made it interactive and use made call instead of the simple text;


js/components/file.js
```diff
@@ -52,16 +52,76 @@ Fliplet.FormBuilder.field('file', {
...
+        return;
+      }
+
+      Fliplet.Media.Files.getAll({
+        files: filesID,
+        fields: ['name', 'url', 'metadata', 'createdAt']
+      }).then(function(files) {
+        $vm.value = files;
+      }).catch(function(err) {
+        console.log(Fliplet.parseError(err));
```
tonytlwu, line 79: @YaroslavOvdii You told me that if `Fliplet.Media.Files.getAll()` is called with multiple file IDs and one or more of the files is not found, the promise is rejected. I asked you to confirm whether the error contains information on which of the files are missing. What was the answer?

@squallstar If it doesn't contain the information on which files (of the provided file IDs) are missing, can that information be provided? We would need that to continue the operation and remove the missing files from the list of files to be displayed (instead of triggering an error alert).

js/components/file.js
```diff
@@ -52,16 +52,76 @@ Fliplet.FormBuilder.field('file', {
...
+    },
     isFileImage: function(file) {
       if (file && file.type) {
         return (file.type.indexOf('image') >= 0);
       }
     },
+    /**
+     * Format bytes as human-readable text.
+     *
+     * @param { Number } bytes size in bytes
```
tonytlwu, line 102:
```suggestion
     * @param {Number} bytes size in bytes
```

tonytlwu (changes requested):

js/components/file.js
```diff
@@ -52,16 +52,76 @@ Fliplet.FormBuilder.field('file', {
   created: function() {
     Fliplet.FormBuilder.on('reset', this.onReset);
   },
+  updated: function() {
+    if (this.readonly) {
+      var $vm = this;
+      var isFileDataLoaded = false;
+      var filesID = _.map(this.value, function(fileURL) {
+        if (typeof fileURL === 'string') {
+          return +fileURL.match(/v1\/media\/files\/([0-9]+)/)[1];
```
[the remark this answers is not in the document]
YaroslavOvdii: @tonytlwu I've tested it through several ways how we can receive previously loaded files.
1) Editing an LFD entry where we have files attached in the data source
2) Set file URLs through the `.field.set()` function
In both cases, we will receive a string with the files' URL. But I agreed that we may enter a different type of URLs with our second way. That is why I'll add one more check to see if the URL is in needed for us format.

js/components/file.js
```diff
@@ -52,16 +52,76 @@ Fliplet.FormBuilder.field('file', {
...
+        }
+
+        isFileDataLoaded = true;
+
+        return null;
+      });
+
+      if (isFileDataLoaded) {
```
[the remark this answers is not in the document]
YaroslavOvdii: @tonytlwu this `if` is needed so we do not load files info in cases: 
1) We already loaded files info
2) We don't have a files ID to load files info

js/components/file.js
YaroslavOvdii: @tonytlwu I've looked into this issue more closely and that what I've discovered:
When we pass not existing ID or `null, NaN, undefined, 0`
Then response will return only existing values.
In that light, I suggest simply remove console.log here, because it wouldn't give any information to the user.

tonytlwu (changes requested):
See unaddressed issue

YaroslavOvdii:
@tonytlwu Sorry missed that, fixed.

tonytlwu (commented):
@sofiiakvasnevska I'd like this tested on development too prior to merging.

tonytlwu (approved):
@sofiiakvasnevska This is ready for testing on staging

Risk: High
Affected: All form users
Verify: App users and custom code is able to control field values as expected

PR #375, opened 2021-03-25
OD-378 [Feature] Upgrade TinyMCE 4 to TinyMCE 5 for Form component

ivandevupp:
## Notes
I added several plugins that are required for correct TinyMCE working in web/native:

- https://github.com/StevenDevooght/tinyMCE-mention/blob/master/mention/plugin.min.js
- https://cdnjs.cloudflare.com/ajax/libs/tinymce/5.7.0/icons/default/icons.min.js
- https://cdnjs.cloudflare.com/ajax/libs/tinymce/5.7.0/plugins/autolink/plugin.min.js
- https://cdnjs.cloudflare.com/ajax/libs/tinymce/5.7.0/plugins/autosave/plugin.min.js
- https://cdnjs.cloudflare.com/ajax/libs/tinymce/5.7.0/plugins/lists/plugin.min.js
- https://cdnjs.cloudflare.com/ajax/libs/tinymce/5.7.0/themes/mobile/theme.min.js
- https://cdnjs.cloudflare.com/ajax/libs/tinymce/5.7.0/skins/content/default/content.min.css
- https://cdnjs.cloudflare.com/ajax/libs/tinymce/5.7.0/skins/ui/oxide-dark/content.mobile.min.css

Can we add them to the assets config?


js/libs/builder.js
```diff
@@ -750,56 +750,6 @@ new Vue({
         this.accessRules.splice(accessRuleIndex, 1);
       }
     },
-    setupCodeEditor: function() {
```
tonytlwu, line 753: This looks like we're removing code editor from somewhere else in the form builder that's unrelated to the Rich Text field. It looks like the form submission confirmation HTML. We can't remove that.

What version of TinyMCE is running in the builder interface? If it's TinyMCE 4, then we can keep this running without having to change anything, right?

tonytlwu:
[The video](https://monosnap.com/file/lwX7YXRHRnWSfDU4NZs7cn1pdtIrEY) at 0:20 showed that the Rich Text editor in the form builder just looks like an ordinary textarea. Once the field is relabelled to something custom, e.g. "Content" or "Bio", then it's indistinguishable from a text area field.

We were OK not having TinyMCE initiated in the page preview in edit mode, but it's important that it's initiated in the form builder interface.

What can we do here? Initialize TinyMCE? If we're stopping the initialization because it's running TinyMCE 4, then maybe our best bet is to run 2 different versions of the initialization config object.

ivandevupp:
@tonytlwu If i got it correctly tinymce should be initialised:
1.  Email templates (v4)
2. Form template (v4)
3. Preview mode (v5)
4. Published/native apps (v5)

And not initialised in edit-preview
image

js/libs/builder.js
ivandevupp: I returned setup for builder.

js/components/wysiwyg.js
```diff
@@ -132,46 +135,71 @@ Fliplet.FormBuilder.field('wysiwyg', {
...
-          'alignleft aligncenter alignright alignjustify | bullist numlist outdent indent',
-          'ltr rtl | link | removeformat code fullscreen'
-        ].join(' | '),
-      image_advtab: true,
-      menubar: false,
+        : ['advlist autolink lists link searchreplace print directionality',
+          'table paste pasteplaintext code'],
       statusbar: false,
+      toolbar: this.readonly || $vm.isInterface
+        ? ['bold italic underline alignleft aligncenter alignright alignjustify'].join(' | ')
```
tonytlwu, line 176: Why would it have a toolbar if it's `readonly`? The previous code suggested that if it's `readonly` then `toolbar: false`.

js/components/wysiwyg.js
```diff
@@ -132,46 +135,71 @@ Fliplet.FormBuilder.field('wysiwyg', {
...
-      autoresize_min_height: lineHeight * this.rows,
-      autofocus: false,
+      convert_urls: false,
       branding: false,
       setup: function(editor) {
         $vm.editor = editor;
 
         editor.on('init', function() {
-          $vm.addPlaceholder();
+          if ($vm.isInterface) {
```
tonytlwu, line 199: If it's not `interface`, e.g. preview/live app, placeholder should also work as it currently does.

js/libs/builder.js
```diff
@@ -983,29 +982,6 @@ new Vue({
         $(selector).removeClass('is-loading');
       }
 
-      $($vm.$refs.templateDescription).tinymce({
```
tonytlwu, line 986: Why was this removed?

tonytlwu (changes requested):
Rejected +1

PR #393, opened 2021-06-11
OD-518 [Fix] - Ensure all Rich Text fields are initialized

skrupnyk:
@sofiiakvasnevska

OD-518 https://weboo.atlassian.net/browse/OD-518

## Description
All Rich text fields are initialised into TinyMCE editors

## Screenshots/screencasts
https://storyxpress.co/video/kps36n3w56zum6z3x

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko @YaroslavOvdii


templates/components/wysiwyg.build.hbs
```diff
@@ -4,7 +4,7 @@
   v-model.trim.lazy="value"
   ref="textarea"
   :name="name"
-  :id="name"
+  :id="tinymceId"
```
tonytlwu, line 7: This tells me that you're using the form component ID as the ID for the WYSIWYG field. This means that if you have 2 WYSIWYG fields in a single form, they would clash as they would have the same ID.

tonytlwu (changes requested):
Rejected +1

templates/components/wysiwyg.build.hbs
skrupnyk: In this case, a unique ID is generated and an ID is added to it by name, this ID is not conflicted, this was provided. This was the only solution that I have found, in order to 'Rich text' has to be initialised into different forms. If there are still options, I can revise and redo.

tonytlwu: This works 👍🏻

tonytlwu (approved):

Risk: High
Affected: Rich text form fields in apps
Verify: Rich text fields are working correctly

PR #404, opened 2021-08-10
OD-544 [Fix] Fix an issue where time picker couldn't be edited

skrupnyk:
@sofiiakvasnevska @inna-bieshulia

OD-544 https://weboo.atlassian.net/browse/OD-544

## Description
Fixed an issue where time picker couldn't be edited

## Screenshots/screencasts
https://storyxpress.co/video/ks7bc6wobsa3qv7v1

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko @S-Volodymyr


js/components/time.js
```diff
@@ -92,27 +92,12 @@ Fliplet.FormBuilder.field('time', {
...
     }
-  },
-  watch: {
-    value: function(val) {
-      if (!val) {
-        this.updateValue(moment().format('HH:mm'));
-      }
+
+    if (!this.value) {
+      this.updateValue(moment().format('HH:mm'));
```
tonytlwu, line 100: It looks like if `this.value` is falsey, it goes through `if (!this.value || this.autofill === 'always') { this.updateValue(moment().format('HH:mm')); this.empty = false; }` once and then `this.updateValue(moment().format('HH:mm'));` again.

Something seems redundant here. Can you explain it had to be coded it this way?

tonytlwu (changes requested):
Solution unclear.

Also, the provided screencast does not show the solution working. In https://storyxpress.co/video/krer4sp6gjoftri4w I see the user editing the form field configuration a couple of times, but no demonstration that the time picker can now be used to change the time value in the app.

Rejected +1

skrupnyk:
> Solution unclear.
> 
> Also, the provided screencast does not show the solution working. In https://storyxpress.co/video/krer4sp6gjoftri4w I see the user editing the form field configuration a couple of times, but no demonstration that the time picker can now be used to change the time value in the app.
> 
> Rejected +1

I pinned a new video

js/components/time.js
skrupnyk: > It looks like if `this.value` is falsey, it goes through `if (!this.value || this.autofill === 'always') { this.updateValue(moment().format('HH:mm')); this.empty = false; }` once and then `this.updateValue(moment().format('HH:mm'));` again.
> 
> Something seems redundant here. Can you explain it had to be coded it this way?

you are right, I corrected the solution and removed the extra code

tonytlwu:
@inna-bieshulia ☝🏻

PR #407, opened 2021-08-17
ID-1171 ID-1095 [Fix] Tidies up time picker UI

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-1171 https://weboo.atlassian.net/browse/ID-1095

Also includes CSS linting

### Builder interface

* Clean up UI so the divider line lines up with the input height

image

### Form render

* Removes duplicated clock icon

image

### DIFF
--- a/css/form.css
+++ b/css/form.css
@@ -1101,12 +1101,12 @@ form hr {
 
 /* Centered increase decrease buttons in the time piker for chrome mac OS */
 
-input[type='time']::-webkit-inner-spin-button,
-input[type='time']::-webkit-outer-spin-button {
+input[type="time"]::-webkit-inner-spin-button,
+input[type="time"]::-webkit-outer-spin-button {
   margin: 1px 0 0 3px;
 }
 
-input[type='time']::-webkit-clear-button {
+input[type="time"]::-webkit-clear-button {
   margin-top: 3px;
 }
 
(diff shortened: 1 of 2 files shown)


css/form.css
```diff
@@ -311,20 +317,20 @@ textarea.form-control {
   display: block;
 }
 
-.checkbox.checkbox-icon input[type="checkbox"] + label {
+.checkbox.checkbox-icon input[type="checkbox"]+label {
```
squallstar, line 320: don't we use spaces in between?

css/form.css
```diff
@@ -367,7 +373,7 @@ html[dir="rtl"] .checkbox.checkbox-icon input[type="checkbox"] + label > span.ch
   transition: all 0.25s ease;
 }
 
-.checkbox.checkbox-icon input[type="checkbox"]:checked + label span.check > i.fa {
+.checkbox.checkbox-icon input[type="checkbox"]:checked+label span.check>i.fa {
```
squallstar, line 376: Pretty sure spaces were meant to stay here. Without it makes it very difficult to read the rule.

squallstar (changes requested):

css/form.css
tonytlwu: @squallstar Corrected. The spaces were being removed by the beautifier's default settings.

squallstar (approved):

PR #410, opened 2021-09-03
OD-558 [Fix] Time/Date picker validation

skrupnyk:
@sofiiakvasnevska @inna-bieshulia

OD-558 https://weboo.atlassian.net/browse/OD-558

## Description
**Problem:** Validation for date picker and time picker works in settings mode.
**Solution:** Changed value handling so that validation works only in preview mode.

## Screenshots/screencasts
https://storyxpress.co/video/kt47xikfs47k5fnfi

## Backward compatibility

Reviewers
@upplabs-alex-levchenko


js/components/date.js
```diff
@@ -95,12 +93,15 @@ Fliplet.FormBuilder.field('date', {
...
+      if (this.datepicker) {
+        this.datePicker.datepicker('setDate', val);
+      }
+
+      if (this.$v.value.$invalid) {
+        this.highlightError();
       }
 
-      this.highlightError();
-      this.$emit('_input', this.name, val, false, true);
```
tonytlwu, line 103: Why are the `false, true` parameters removed? They were added as part of https://github.com/Fliplet/fliplet-widget-form-builder/pull/409

tonytlwu (approved):

tonytlwu (changes requested):
Needs clarification

js/components/date.js
upplabs-alex-levchenko: We've removed 'updateValue' usages which potentially were calling a loop of 'emit' events and now we are firing one 'emit' after the component load and one each time after value is updated.
Should we ask QA to check the issue for you previous fix? If yes, please let me know which one :)

tonytlwu: The bug was https://weboo.atlassian.net/browse/ID-1394 Please make sure QA verifies the bug is not re-introduced.

skrupnyk:
new PR https://github.com/Fliplet/fliplet-widget-form-builder/pull/412

PR #413, opened 2021-09-17
OD-542 [Feat] Add columns to data source using the button "Create columns"

OleksandrLavrynenko:
@sofiiakvasnevska

OD-542 https://weboo.atlassian.net/browse/OD-542

## Description
DataSource columns are no longer added automatically. To create the desired columns in the data sources, the user must add them manually or using the "Create columns" button in the form settings

## Screenshots/screencasts
https://storyxpress.co/video/ktoehutytl822koze

## Backward compatibility
This change is fully backward compatible.

## Reviewers
@upplabs-alex-levchenko

PR #419, opened 2021-10-19
ID-1547 [Fix] Ensure data source hooks are kept up to date when saving forms

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-1547


tonytlwu:
@squallstar The management of data source hooks and new columns used to both happen at the same time when the form instance is saved. This is no longer the case.

With this change and [this recent change](https://github.com/Fliplet/fliplet-widget-form-builder/pull/413), data source hooks are always updated (as it was) when the form instance is saved. New data source columns are only added on demand when user clicks on an **Add missing data source columns** button. 

In my opinion it means that they will never be called together and it's required for these requests to be separate.

Let me know if I've missed anything.

PR #424, opened 2021-11-04
OD-637 [Fix] Ensures checkboxes radio buttons and signatures are restored when loading data

OleksandrLavrynenko:
@romanyosyfiv

OD-637 https://weboo.atlassian.net/browse/OD-637

## Description
**Problem**: The data for the checkbox and radio button options was not generated in the checkbox and radio button components and this caused an error with the display of the selected options. The signature was not displayed for editing because there was no functionality to display it.
**Solution**: The settings of the selected options have been transferred to the checkbox component and the radio button component. The preview is implemented for the signature component

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


interface.html
```diff
@@ -177,7 +177,7 @@ <h3>Submission</h3>
                 </div>
                 <div v-if="settings.autobindProfileEditing">
                   <p class="help-block">
-                    <small>If the form loads data from user’s profile, it can only be used to update the user’s profile.</small>
+                    <small>If the form loads data from user’s profile, it can only be used to ypdate the user’s profile.</small>
```
tonytlwu, line 180:
```suggestion
                    <small>If the form loads data from user’s profile, it can only be used to update the user’s profile.</small>
```

js/components/checkbox.js
```diff
@@ -65,6 +65,24 @@ Fliplet.FormBuilder.field('checkbox', {
     }
   },
   created: function() {
+    var $vm = this;
+
+    if (this.value.length > 0) {
+      var selectedOptions = [];
+
+      this.value.forEach(function(element) {
```
tonytlwu, line 73:
```suggestion
      this.value.forEach(function(value) {
```

js/components/checkbox.js
```diff
@@ -65,6 +65,24 @@ Fliplet.FormBuilder.field('checkbox', {
...
+        var selectedOption = _.find($vm.options, function(option) {
+          return option.label === element || option.id === element;
```
tonytlwu, line 75: Sometimes an option could have `label` only. Sometimes it could have `label` and `id`. When it has both `label` and `id`, then it **MUST** match the `id`, not the `label`.

If the value is `foo` and the options are:

```js
[
  { label: "foo", id: "bar" },
  { label: "foo2", id: "foo" }
]
```

The correct mat would be `{ label: "foo2", id: "foo" }` because the `id` matches. However, the current implementation has a flaw and will match with `{ label: "foo", id: "bar" }` instead. I recommend something like this:

```suggestion
          return ('label' in option && 'id' in option) ? option.id === value : option.label === value;
```

js/components/radio.js
```diff
@@ -45,5 +45,14 @@ Fliplet.FormBuilder.field('radio', {
       this.$refs.radioButton[newIndex].focus();
       this.clickHandler(this.options[newIndex]);
     }
+  },
+  created: function() {
+    var $vm = this;
+
+    var selectedOption = _.find($vm.options, function(option) {
+      return option.label === $vm.value || option.id === $vm.value;
```
tonytlwu, line 53: Same problem as the checkbox field.

js/libs/form.js
```diff
@@ -211,43 +211,13 @@ Fliplet.Widget.instance('form-builder', function(data) {
               }
               break;
 
-            case 'flCheckbox':
```
tonytlwu, line 214: You mentioned in the description:

> **Problem**: The data for the checkbox and radio button options was not generated in the checkbox and radio button components and this caused an error with the display of the selected options. The signature was not displayed for editing because there was no functionality to display it.
> **Solution**: The settings of the selected options have been transferred to the checkbox component and the radio button component. The preview is implemented for the signature component

I don't understand why the data wasn't being generated, and there why the problem must be solved by transferring the option setting to the component. I assume this worked at one point when the feature was initially built.

@OleksandrLavrynenko You may want to address this first before tackling the other changes.

tonytlwu (changes requested):

js/libs/form.js
```diff
@@ -211,43 +211,13 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
-              if (fieldData.length > 0) {
-                var inOptions = [];
-
-                fieldData.forEach(function (element) {
-                  var match = _.find(field.options, function (option) {
```
OleksandrLavrynenko, line 219: @tonytlwu This code cannot be executed because the array field.options is empty unless it has been changed in the widget settings. So I decided to run this code in the checkbox component, because it always has relevant options.

tonytlwu: @OleksandrLavrynenko In that case, can you leave the following code so that developers can follow the thread in the future:

```js
case 'flCheckbox':
case 'flRadio':
  // Add comments to explain where to find the code for setting the field value and why it needs to be elsewhere
  break;
```

js/libs/form.js
```diff
@@ -211,42 +211,15 @@ Fliplet.Widget.instance('form-builder', function(data) {
...
-              // Work only if passed value is a string
-              // Altered check to support old version of the form builder and if a user provides data as ID not a Label it will work correctly
-              var match = _.find(field.options, function (option) {
-                return option.label === fieldData || option.id === fieldData;
-              });
-
-              field.value = match ? fieldData : '';
-              break;
+            case 'flCheckbox': // There is no validation and value assignment as there is no access to checkbox options. This is implemented in the checkbox component.
+            case 'flRadio': // There is no validation and value assignment as there is no access to radio options. This is implemented in the radio component.
```
tonytlwu, line 222: @OleksandrLavrynenko Currently this says checkbox and radio will also execute this line: `field.value = fieldData;` Is that intentional or do we want to stop the code execution?

```suggestion
            case 'flRadio': // There is no validation and value assignment as there is no access to radio options. This is implemented in the radio component.
              break;
```

OleksandrLavrynenko: > Is that intentional

**@tonytlwu**  This is intentional because we need to get the values

tonytlwu (approved):
@romanyosyfiv @inna-bieshulia This is ready for testing

Risk: High
Affected: Checkbox, radio and signature fields
Verify: The affected fields can save, submit and load data correctly

PR #425, opened 2021-11-08
OD-642 [Fix] Remove duplicated 'description' fields for image, file, signature and star rating fields

skrupnyk:
@romanyosyfiv

OD-642 https://weboo.atlassian.net/browse/OD-642

## Description
**Problem:** There is a double call to the "description" field. This happens in the component file and in the core file.
**Solution:** One of the calls was deleted. The description field is left only in the component file.

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/libs/core.js
```diff
@@ -364,7 +364,7 @@ Fliplet.FormBuilder = (function() {
 
       component.props._componentsWithDescription = {
         type: Array,
-        default: ['flInput', 'flCheckbox', 'flRadio', 'flEmail', 'flNumber', 'flTelephone', 'flUrl', 'flTextarea', 'flWysiwyg', 'flSelect', 'flDate', 'flTime', 'flStarRating', 'flSignature', 'flImage', 'flFile']
```
tonytlwu, line 367: Why is it that the descriptions needed to be called in the component files as well? By removing them from the list, it looks like those 4 fields are not "componentsWithDescription", but they are -- it would cause future confusion and possibility of regression.

tonytlwu (approved):
@romanyosyfiv @inna-bieshulia This is ready for testing

Risk: Medium
Affected: Studio users configuring form components
Verify: File, Image, Signature and Star rating fields have the correct configuration fields

PR #467, opened 2022-11-08
XOR-303 [FEAT] Add Slider field to form component

NikitaJadhavXoriant:
### Product areas affected

Widgets -> Add Form -> Advanced Section -> Select Slider field 

### What does this PR do?

Implemented code to add new Slider field in form

### JIRA ticket

[JIRA](https://weboo.atlassian.net/browse/XOR-303)

### Result

https://user-images.githubusercontent.com/108272606/200498813-df37f9ce-29ba-415c-a100-cdfb3d0bf67e.mp4

### Checklist

None

### Deployment instructions

None

### Author concerns

None


css/builder.css
```diff
@@ -556,6 +556,34 @@ input[type="time"]::-webkit-calendar-picker-indicator {
   opacity: 0;
 }
 
+/* SLIDER FIELD */
+
+.form-group.fl-range-slider .range-slider .range-slider-handle,
```
tonytlwu, line 561: Why are these necessary? It looks like it's just overwriting the same styles that are already defined in `public/assets/fliplet-ui-rangeslider/1.0/ui.rangeslider.css` of https://github.com/Fliplet/fliplet-api/pull/5566/files

We should only be including anything that's uniquely different from the JS API.

css/form.css
```diff
@@ -1031,6 +1031,51 @@ html[dir="rtl"] .radio.radio-icon input[type="radio"] + label > span.check {
   color: #e1e1e1;
 }
 
+/* SLIDER FIELD */
+
+[data-widget-package="com.fliplet.form-builder"] .form-group.fl-range-slider .range-slider {
```
tonytlwu, line 1036: Same question as `css/builder.css`

templates/configurations/slider.interface.hbs
```diff
@@ -0,0 +1,4 @@
+<div v-show="defaultValueSource === 'default'" class="form-group">
+  <label>Default value</label>
+  <input class="form-control" v-model.trim="value" placeholder="Default value"></input>
```
tonytlwu, line 3:
```suggestion
  <input class="form-control" v-model.trim="value" placeholder="Default value">
```

templates/configurations/form.interface.hbs
```diff
@@ -95,6 +95,25 @@
       </div>
     </div>
 
+    {{#isSlider}}
+    <div class="form-group cleafix">
+      <div class="col-xs-12">
+        <label>Minimum Value Number</label>
+        <input class="form-control" type="text" v-model.trim="min" placeholder="Default: 0"/>
```
tonytlwu, line 102: Let's use `number` for min, max and step.

```suggestion
        <input class="form-control" type="number" v-model.trim="min" placeholder="Default: 0"/>
```

js/libs/form.js
```diff
@@ -225,6 +225,11 @@ Fliplet().then(function () {
               case 'flCheckbox': // There is no validation and value assignment as there is no access to checkbox options. This is implemented in the checkbox component.
               case 'flRadio': // There is no validation and value assignment as there is no access to radio options. This is implemented in the radio component.
 
+              case 'flSlider':
```
tonytlwu, line 228: I don't understand why this is necessary or why it's here.

This entire `switch` clause is about assigning `fieldData` from the data source to `field.value`.

Progress is reassigned back to the field in the `if (progress && !isEditMode)` part after this `switch` clause.

js/components/slider.js
```diff
@@ -0,0 +1,87 @@
+Fliplet.FormBuilder.field('slider', {
+  name: 'Slider',
+  category: 'Advanced',
+  props: {
+    placeholder: {
```
tonytlwu, line 5: There's no `placeholder` for the slider field. Let's remove this. Looks like there's nothing referencing it anyway.

```suggestion
```

tonytlwu (changes requested):

templates/components/slider.build.hbs
```diff
@@ -0,0 +1,14 @@
+<div v-if="description" class="help-block description">\{{ description }}</div>
+<div class="form-group fl-range-slider" :class="{ 'readonly': readonly }" ref="Slider">
```
tonytlwu, line 2: Use _camelCase_

```suggestion
<div class="form-group fl-range-slider" :class="{ 'readonly': readonly }" ref="slider">
```

css/builder.css
```diff
@@ -556,6 +556,20 @@ input[type="time"]::-webkit-calendar-picker-indicator {
...
+[data-widget-package="com.fliplet.form-builder"] .form-group.fl-range-slider.readonly .range-slider-handle-value::before {
+  border-top-color: #e1e1e1;
+}
+
+[data-widget-package="com.fliplet.form-builder"] .form-group.fl-range-slider .range-slider.range-slider-active .range-slider-handle {
+  box-shadow: 0px 0px 0px 2px rgb(0 171 209 / 25%);
+}
+
+.slider-input {
+  width: 100% !important;
```
tonytlwu, line 570: Why is this necessary?

tonytlwu (changes requested):

css/builder.css
NikitaJadhavXoriant: Hi @tonytlwu I added this code because while onSlide screen is producing glitch as slider input i having width auto, so its producing scrollbar while doing onslide.

tonytlwu: image

@NikitaJadhavXoriant See above. [The prototype](https://codepen.io/tonytlwu/pen/QWrzERN) showed that the slider range library should give it styles such that it has 1 x 1 size and is absolutely positioned. What's causing this to behave the way you described and have a different width? If anything, the width override should be `1px`?

And if you really must use a style override, you should use a selector that won't stop other users from using a `slider-input` class, which is pretty generic, e.g. `[data-widget-package="com.fliplet.form-builder"] .form-group.fl-range-slider > input`.

NikitaJadhavXoriant: Hi @tonytlwu please review this [comment](https://weboo.atlassian.net/browse/XOR-303?focusedCommentId=39050).

templates/configurations/slider.interface.hbs
NikitaJadhavXoriant: did the changes.

tonytlwu (approved):

templates/configurations/slider.interface.hbs
tonytlwu: Doesn't look like this was resolved. Where did you make the changes?

NikitaJadhavXoriant: Hi @tonytlwu, while fixing PR other feedback, accidentally it got reverted.

tonytlwu (approved):

PR #477, opened 2022-12-28
XOR-718 [FIX] On clear reset field to Default value

NikitaJadhavXoriant:
**Product areas affected**

Studio -> Form -> Clear

**What does this PR do?**

Implemented code for clear (reset) button to reset field with its default value.

**JIRA ticket**

[click here](https://weboo.atlassian.net/browse/XOR-718)

**Result**

https://user-images.githubusercontent.com/108272606/209801421-cfe08357-0962-4c1b-bcef-87ba7e1d9643.mp4

**Checklist**

None

**Testing instructions**

None

**Deployment instructions**

None

**Author concerns**

None

[XOR-718]: https://weboo.atlassian.net/browse/XOR-718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yogeshmsharma-xoriant (approved):
ok

js/libs/form.js
```diff
@@ -300,12 +303,17 @@ Fliplet().then(function () {
           }
 
           var $vm = this;
+          var entryLoaded = false;
+
+          if (typeof trackEvents === 'undefined') {
```
tonytlwu, line 308: We shouldn't use the presence of `trackEvents`, which is only used to dictate whether the reset should be tracked, to signify whether this "is a reset action".

The `isReset` flag also isn't entirely to future readers why this is necessary because the function is called `reset()`. We need better explanations and naming.

tonytlwu (changes requested):

js/libs/form.js
tonytlwu: @yogeshmsharma-xoriant I suggest refactoring to use functions like this:

```js
resetForm() {
},
submit() {
  // Submit the form as it does now

  // Reset form
  resetForm()
},
reset() {
  // Reset form
  resetForm()
}
```

This way, you can have code that runs in `reset()` that isn't in `resetForm()`. `reset()` is what happens when the reset is intentional whereas `resetForm()` is the internal function for resetting the form.

yogeshmsharma-xoriant (approved):
looks good to me

js/libs/form.js
```diff
@@ -291,7 +294,11 @@ Fliplet().then(function () {
           }
           this.isSent = false;
         },
-        reset: function(trackEvents) {
+        reset: function () {
+            isResetAction = true;
```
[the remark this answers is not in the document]
NikitaJadhavXoriant: Hi @tonytlwu,

We are setting isResetAction flag initially to false, and on click of clear button we are setting isResetAction flag to true. 

If form has both image and file component or mulitple images/files, in this case we dont want to reset flag to false because[ switch case](https://github.com/Fliplet/fliplet-widget-form-builder/blob/master/js/libs/form.js#L195-L206) will not execute for next image / file fields.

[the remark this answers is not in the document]
tonytlwu: @NikitaJadhavXoriant This looks like if `reset()` is run, `isResetAction` is changed from `false` to `true`. The next time `submit()` is run (which calls `resetForm()`, the `isResetAction` variable will still be `true` -- I don't think that's what we want, right?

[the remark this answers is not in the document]
NikitaJadhavXoriant: @tonytlwu Agree, missed this scenario. Adding isResetAction to false in onSubmit method.

yogeshmsharma-xoriant (commented):
@NikitaJadhavXoriant  hope you have tested all below scenarios

1. Clear and then submit 
2. Submit 
3. Clicking clear twice 
4.clcking submit twice 
5. Form reload and then direct submit 
6. form reload and clear 
7.  Test with old forms and new form

NikitaJadhavXoriant:
> @NikitaJadhavXoriant hope you have tested all below scenarios
> 
> 1. Clear and then submit
> 2. Submit
> 3. Clicking clear twice
>    4.clcking submit twice
> 4. Form reload and then direct submit
> 5. form reload and clear
> 6. Test with old forms and new form

@yogeshmsharma-xoriant  done

yogeshmsharma-xoriant (dismissed):
approved

js/libs/form.js
```diff
@@ -453,6 +466,8 @@ Fliplet().then(function () {
             action: 'submit'
           });
 
+          isResetAction = false;
```
tonytlwu, line 469: Will it make sense to set `isResetAction` to `false` at the end of `resetForm()` instead?

tonytlwu (changes requested):
@yogeshmsharma-xoriant Does not meet coding standards

js/libs/form.js
NikitaJadhavXoriant: Hi @tonytlwu 

We can't do isResetAction = false in resetForm, due to parallel method execution, please refer [this](https://weboo.atlassian.net/browse/XOR-718?focusedCommentId=40251) comment for your reference

cc @yogeshmsharma-xoriant

tonytlwu: @NikitaJadhavXoriant Then perhaps `isResetAction` should be set to `false` in the `submit()` function only right before calling `resetForm()`?

yogeshmsharma-xoriant (approved):
go ahead

tonytlwu (changes requested):
Does not meet coding standards

yogeshmsharma-xoriant (approved):
eslint changes looks ok

tonytlwu (changes requested):
Does not meet coding standards

yogeshmsharma-xoriant (approved):
looks ok for  now

tonytlwu (approved):

PR #480, opened 2023-03-15
XOR-837 [FIX] Ensure warning message is shown for empty row and column option of the matrix field

NikitaJadhavXoriant:
**Product areas affected**

Studio -> Form -> Matrix field

**What does this PR do?**

Implemented code to display warning message when empty row and column option is given for matrix field.

**JIRA ticket**

[click here](https://weboo.atlassian.net/browse/XOR-837)

**Result**

https://user-images.githubusercontent.com/108272606/225288050-738ca6f1-7d4d-483f-a6f9-1440b832b029.mp4

**Checklist**

None

**Testing instructions**

None

**Deployment instructions**

None

**Author concerns**

None


yogeshmsharma-xoriant (dismissed):
looks ok

js/libs/core.js
```diff
@@ -502,6 +502,21 @@ Fliplet.FormBuilder = (function() {
 
             break;
 
+          case 'flMatrix':
+            if (this.columnOptions.length <= 0) {
```
tonytlwu, line 506: The length can't be less than 0. 

We also shouldn't allow the following scenarios:

- spaces only
- line breaks only
- line breaks and spaces only

This means the row and column options need to be trimmed and cleaned up before validation and before saving.

```suggestion
            if (this.columnOptions.length === 0) {
```

tonytlwu (changes requested):

js/libs/core.js
NikitaJadhavXoriant: @tonytlwu ,

line breaks and spaces scenarios we are handling already [here](https://github.com/Fliplet/fliplet-widget-form-builder/blob/master/js/libs/core.js#L724).

=== change done.

yogeshmsharma-xoriant (approved):

tonytlwu (approved):

PR #490, opened 2023-04-12
ID-2984 [FEAT] Created date range field for form component

armine-fliplet:


js/components/dateRange.js
```diff
@@ -0,0 +1,280 @@
+Fliplet.FormBuilder.field('dateRange', {
+  name: 'Date range',
+  category: 'Text inputs',
```
tonytlwu, line 3: We should put this field along with the Date and Time fields into a new category called **Date & time**, after the **Multiple options** category.

js/components/dateRange.js
```diff
@@ -0,0 +1,280 @@
...
+    },
+    autofill: {
+      type: String,
+      default: 'always'
+    },
+    defaultSource: {
+      type: String,
+      default: 'submission'
+    },
+    defaultEndDays: {
```
tonytlwu, line 23:
```suggestion
    defaultRangeDuration: {
```

js/components/dateRange.js
```diff
@@ -0,0 +1,280 @@
...
+
+    if (this.autofill === 'default' || this.autofill === 'always') {
+      this.value = {
+        start: this.today,
+        end: this.today
+      };
+      this.empty = false;
+    }
+
+    if (this.autofill === 'empty') {
```
tonytlwu, line 65: Join this with the previous line.

```suggestion
    else if (this.autofill === 'empty') {
```

Or use `switch (this.autofill)`

js/components/dateRange.js
```diff
@@ -0,0 +1,280 @@
...
+      this.value = {
```
tonytlwu, line 66:
```suggestion
      this.value = null;
```

js/components/dateRange.js
```diff
@@ -0,0 +1,280 @@
...
+    return rules;
+  },
+  computed: {
+    isApplyCurrentDateField: function() {
+      return this.autofill === 'always' || this.autofill === 'default';
+    },
+    predefinedRanges: function() {
+      return [
+        {
+          label: T('widgets.form.dateRange.predefinedRanges.today')
```
tonytlwu, line 95: You haven't defined these translation keys in `translation.json`. See https://docs.google.com/document/d/1gMBCtU71FnQPInDxg8HJNBgDzfzUTpBH7s3a1e8H-HI/edit#heading=h.xiit943bw330

Also, I would separate `{ label, value }`, e.g.

```
{ label: T('widgets.form.dateRange.predefinedRanges.today'), value: 'today' }

// or

{ label: T('widgets.form.dateRange.predefinedRanges.nextMonth'), value: 'nextMonth' }
```

This way the `label` is localised and translated, but `value` can remain the same.

js/components/dateRange.js
```diff
@@ -0,0 +1,280 @@
...
+      rules.value.required = window.validators.required;
+    }
+
+    return rules;
+  },
+  computed: {
+    isApplyCurrentDateField: function() {
+      return this.autofill === 'always' || this.autofill === 'default';
+    },
+    predefinedRanges: function() {
```
tonytlwu, line 92: I don't think this needs to be a computed property. It should be fine as a `data` property.

js/components/dateRange.js
```diff
@@ -0,0 +1,280 @@
...
+      this.$emit('_input', this.name, val, false, true);
+    },
+    selectedRange: function(val) {
+      var newDate = {
+        start: '',
+        end: ''
+      };
+
+      switch (val) {
+        case T('widgets.form.dateRange.predefinedRanges.today'):
```
tonytlwu, line 152: No need to check against localised string if you follow the `{ label, value }` structure I showed above.

You can then just use `newDate = this.getDate(val)`

js/components/dateRange.js
```diff
@@ -0,0 +1,280 @@
...
+        this.highlightError();
+      }
+
+      if (this.dateRange) {
+        this.dateRange.set(val, true);
+      }
+
+      this.$emit('_input', this.name, val, false, true);
+    },
+    selectedRange: function(val) {
```
tonytlwu, line 145:
```suggestion
    selectedRange: function(range) {
```

js/components/dateRange.js
```diff
@@ -0,0 +1,280 @@
...
+
+        $vm.value = value;
+        $vm.updateValue();
+      });
+    },
+    getDate: function(option) {
+      switch (option) {
+        case 'today':
+          return {
+            start: moment().locale('en').format('YYYY-MM-DD'),
```
tonytlwu, line 227: Given how often `.locale('en').format('YYYY-MM-DD')` is repeated, can we make something more reusable?

js/components/dateRange.js
```diff
@@ -0,0 +1,280 @@
...
+          };
+        default:
+          break;
+      }
+    },
+    onBeforeSubmit: function(data) {
+      // Empty date fields are validated to null before this hook is called
+      if (this.autofill === 'always' && data[this.name] === null) {
+        data[this.name] = this.defaultSource === 'submission' ?
+          {
```
tonytlwu, line 268: Put the `?` and `:` in the new line

```suggestion
          ? {
```

templates/components/dateRange.build.hbs
```diff
@@ -0,0 +1,46 @@
...
+  <label
+    v-if="showPredefinedRanges"
+    :for="name"
+    class="select-proxy-display"
+    :class="{ 'input-focused': isInputFocused , 'readonly': readonly }"
+  >
+    <select
+      class="form-control hidden-select focus-outline"
+      name="predefinedRangesOptions"
+      id="predefinedRangesOptions"
```
tonytlwu, line 11: https://docs.google.com/document/d/1XPlQHPbqp3rGGTekcC05bKtwKDDjrU2RMJ4VIHLXp64/edit#heading=h.yewnp98kurqg

Don't use non-unique IDs in apps because there could be multiple fields and instances of the widget. Personally I don't think `name` or `id` is required here.

templates/components/dateRange.build.hbs
```diff
@@ -0,0 +1,46 @@
...
+      v-model.lazy="selectedRange"
+      v-on:keydown.prevent="readonly ? false : true"
+      tabindex="0"
+      @blur="onBlur()"
+    >
+      <option v-if="placeholder" value="">\{{ placeholder }}</option>
```
tonytlwu, line 17: If predefined options need to be shown, then this option should always be available. It shouldn't have anything to do with whether there's a truthy value for `placeholder`.

js/components/dateRange.js
```diff
@@ -0,0 +1,280 @@
...
+      this.dateRange = Fliplet.UI.DateRange(this.$refs.dateRange, {
+        required: this.required || this.autofill === 'always',
+        forceRequire: false,
+        value: this.value,
+        readonly: this.readonly
+      });
+
+      this.dateRange.change(function(value) {
+        if (!value || !value.start || !value.end) {
+          this.selectedRange = '';
```
tonytlwu, line 216: `this.selectedRange` should be set to `''` whenever the value changes after the range is moved away from a predefined range.

tonytlwu (changes requested):

css/builder.css
```diff
@@ -556,6 +556,39 @@ input[type="time"]:not(.basic-time-picker)::-webkit-calendar-picker-indicator {
...
+  margin-top: 12px;
+  flex-wrap: nowrap;
+}
+
+.fl-date-range .form-group {
+  width: 100%;
+  min-width: 156px;
+}
+
+.fl-date-range .arrow-right {
```
tonytlwu, line 571: The arrow should be part of the JS API styling. The form component won't need to apply further styling unless it's styling specific to the form component. We want everyone else using the JS API to have the right styles including the arrow.

js/components/dateRange.js
```diff
@@ -0,0 +1,239 @@
...
+            end: this.getLocaleDate(moment().subtract(1, 'days'))
+          };
+        default:
+          return {
+            start: this.getLocaleDate(moment()),
+            end: this.getLocaleDate(moment())
+          };
+      }
+    },
+    getLocaleDate: function(date) {
```
tonytlwu, line 226:
```suggestion
    formatLocaleDate: function(date) {
```

js/components/dateRange.js
```diff
@@ -0,0 +1,239 @@
...
+        return '-- ' + T('widgets.form.dateRange.rangePlaceholder');
+      }
+    }
+  },
+  data: function() {
+    return {
+      dateRange: null,
+      isInputFocused: false,
+      isPreview: Fliplet.Env.get('preview'),
+      today: moment().locale('en').format('YYYY-MM-DD'),
```
tonytlwu, line 43:
```suggestion
      today: this.formatLocaleDate(moment()),
```

js/libs/templates.js
```diff
@@ -8,6 +8,11 @@ var systemTemplates = [{
       name: 'Question 1',
       label: 'Enter your first question'
     },
+    {
```
tonytlwu, line 11: We don't want the initial form template to include a date field

tonytlwu (changes requested):
There are conflicts in the PR that you need to resolve too

css/builder.css
```diff
@@ -551,6 +551,18 @@ input[type="time"]:not(.basic-time-picker)::-webkit-calendar-picker-indicator {
   opacity: 0;
 }
 
+.fl-date-range {
```
tonytlwu, line 554: Could these have been in the JS API too? The question is: can the JS API UI work/render as expected without these CSS?

js/components/dateRange.js
```diff
@@ -0,0 +1,239 @@
+Fliplet.FormBuilder.field('dateRange', {
+  name: 'Date range',
+  category: 'Date & time',
+  props: {
+    value: {
+      type: Object,
+      default: {
```
tonytlwu, line 7: Could we start the value as `null`, given the the date range JS API's getter returns either a valid date range object or `null`.

css/builder.css
armine-fliplet: Yes, JS API can render the UI

tonytlwu (approved):

PR #491, opened 2023-04-18
ID-2977 [FEAT] Implement time range form field

armine-fliplet:


js/components/timeRange.js
```diff
@@ -0,0 +1,144 @@
+Fliplet.FormBuilder.field('timeRange', {
+  name: 'Time range',
+  category: 'Text inputs',
```
tonytlwu, line 3: Assign this to the **Date & time** category

tonytlwu (changes requested):

js/libs/core.js
```diff
@@ -155,6 +155,7 @@ Fliplet.FormBuilder = (function() {
         });
       };
 
+
```
tonytlwu, line 158:
```suggestion
```

templates/configurations/timeRange.interface.hbs
```diff
@@ -41,5 +47,12 @@
 <div v-else class="form-group">
   <label>Default value</label>
   <div class="text-danger" v-if="errors.requiredField"> \{{ errors.requiredField }} </div>
-  <input type="time" class="form-control basic-time-picker" v-model.trim="value" />
+  <div class="fl-time-range form-html">
```
tonytlwu, line 50: This doesn't look right to me. https://github.com/Fliplet/fliplet-api/pull/5725/files#diff-4612bda789533f214313c060a4fa54e08258786d64a0f1355e6bcb7a64d96129R13-R29 says a different markup is needed. Can you also share a screenshot of this one it's correctly done?

armine-fliplet: image

this is for the custom option

tonytlwu: @armine-fliplet Then let's use the same UI markup you've defined for `Fliplet.UI.TimeRange`. It'll ensure consistent look and feel with the time range field as well.

tonytlwu (approved):

PR #495, opened 2023-04-28
ID-2973 [FEAT] Implement Timer form field

armine-fliplet:


templates/configurations/form.interface.hbs
```diff
@@ -108,6 +108,28 @@
...
+    {{#isTimer}}
+      <div class="form-group">
+        <div class="col-xs-12">
+          <label class="control-label">Timer type</label>
+        </div>
+        <div class="col-xs-12">
+          <div class="radio radio-icon">
+            <input type="radio" id="stopwatch" name="stopwatch" value="stopwatch" v-model="type">
+            <label for="stopwatch">
+              <span class="check"><i class="fa fa-circle"></i></span> \{{ T('widgets.form.timer.types.stopwatch') }}
```
tonytlwu, line 120: Studio interface doesn't need localisation.

tonytlwu (changes requested):

js/components/timer.js
```diff
@@ -0,0 +1,194 @@
...
+
+    if (this.required && !this.readonly) {
+      rules.value.required = window.validators.required;
+    }
+
+    return rules;
+  },
+  computed: {
+    stringValue: function() {
+      return `${this.hours < 10 ? '0' + this.hours : this.hours} : ${this.minutes < 10 ? '0' + this.minutes : this.minutes} : ${this.seconds < 10 ? '0' + this.seconds : this.seconds}`;
```
tonytlwu, line 51: I used this prompt in ChatGPT

```
I have a value that represents the number of seconds, which doesn't have a upper limit. How do I use JavaScript to return a string that is in the format of `HH:mm:ss`, representing the value in terms of hours, minutes and seconds?
```

I got this function back, which I think contains the logic you can use for `setDisplayValues` and computing `stringValue`. The logic is simpler and clearer to read.

```js
function formatSeconds(seconds) {
  var hours = Math.floor(seconds / 3600);
  var minutes = Math.floor((seconds % 3600) / 60);
  var remainingSeconds = seconds % 60;

  // Add leading zeros if necessary
  var hoursStr = hours.toString().padStart(2, '0');
  var minutesStr = minutes.toString().padStart(2, '0');
  var secondsStr = remainingSeconds.toString().padStart(2, '0');

  return hoursStr + ':' + minutesStr + ':' + secondsStr;
}
```

js/components/timer.js
```diff
@@ -0,0 +1,194 @@
...
+      this.setValueFromDefaultSettings({ source: this.defaultValueSource, key: this.defaultValueKey });
+    }
+
+    if (this.type === 'timer') {
+      this.initialValue = this.toSeconds(this.hours, this.minutes, this.seconds);
+    } else {
+      this.initialValue = 0;
+    }
+
+    await Fliplet.App.Storage.get(this.name).then(function(value) {
```
tonytlwu, line 178: Why is `await` needed if we're not returning the result?

```suggestion
   Fliplet.App.Storage.get(this.name).then(function(value) {
```

js/components/timer.js
```diff
@@ -0,0 +1,194 @@
...
+      this.value = this.type === 'timer' ? Math.min(this.initialValue, data) : Math.max(data, 0);
+      this.setDisplayValues(this.value);
+
+      if (this.timerStatus === 'running') {
+        this.startTimestamp = moment().unix();
+        Fliplet.App.Storage.set(this.name, this.startTimestamp);
+      }
+    },
+    updateValue() {
+      return Math.floor(this.value) + (moment().unix() - this.startTimestamp);
```
tonytlwu, line 162: I don't think we need to round this. If you do, let me know why.

```suggestion
      return this.value + (moment().unix() - this.startTimestamp);
```

js/components/timer.js
```diff
@@ -0,0 +1,194 @@
...
+      return this.updateValue();
+    },
+    set: function(data) {
+      if (!/^(\d+.)*(\d+)$/.test(data)) {
+        return;
+      }
+
+      data = parseFloat(data).toFixed(3);
+
+      this.value = this.type === 'timer' ? Math.min(this.initialValue, data) : Math.max(data, 0);
```
tonytlwu, line 153:
```suggestion
      this.value = this.type === 'timer'
        ? Math.min(this.initialValue, data)
        : Math.max(data, 0);
```

tonytlwu: Also, if it's a timer, you should still use Math.max to make sure the value is bigger than 0

js/components/timer.js
```diff
@@ -0,0 +1,194 @@
...
+      default: 0
+    }
+  },
+  data: function() {
+    return {
+      isPreview: Fliplet.Env.get('preview'),
+      timerStatus: 'stopped',
+      initialValue: 0,
+      timerInterval: null,
+      startTimestamp: moment().unix()
```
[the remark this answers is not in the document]
armine-fliplet: @tonytlwu should I use `moment().valueOf() / 1000` in start timer too?

css/builder.css
```diff
@@ -556,7 +556,15 @@ input[type="time"]:not(.basic-time-picker)::-webkit-calendar-picker-indicator {
...
 
 .fl-timer {
-  column-gap: 8px;
+  width: 100%;
+  display: flex;
+  column-gap: 10px;
+  margin-top: 5px;
+}
+
+.fl-timer > div{
```
tonytlwu, line 565: Your IDE should be able to autofix this error when you save the file. Can you confirm your setup is working correctly?

```suggestion
.fl-timer > div {
```

tonytlwu (changes requested):

js/components/timer.js
```diff
@@ -184,17 +187,19 @@ Fliplet.FormBuilder.field('timer', {
...
     });
 
     this.stringValue = this.formatSeconds(this.initialTimerValue);
 
     if (this.autostart) {
-      this.start();
+      this.status = 'running';
     }
 
+    $(this.$refs.timer).data('flTimer', this);
```
tonytlwu, line 201: We don't need this any more.

```suggestion
```

js/libs/form.js
```diff
@@ -1095,6 +1095,7 @@ Fliplet().then(function() {
             },
             field: function(key) {
               var field = $form.getField(key);
+              var instance = _.find($form.$children, { name: field.name });
```
tonytlwu, line 1098: If you look the line after this, you'll find that we can't assume `field` is retrieved from the line above, so you can't assume `field.name` is valid.

You should also add a check to make sure the instance is found, or return an error (with the same message as below).

Also, let's define this as `$field` to imply it's a Vue object representing the field.

Then, update https://github.com/Fliplet/fliplet-widget-form-builder/blob/a5a3763524b670471682a896d213758838772a5a/js/libs/form.js#L1074 to also include `$instance: $form`.

js/libs/form.js
```diff
@@ -1276,7 +1277,44 @@ Fliplet().then(function() {
...
+                  }
+
+                  eventListeners[field.name] = eventListeners[field.name] || {};
+                  eventListeners[field.name][eventName] = eventListeners[field.name][eventName] || [];
+                  eventListeners[field.name][eventName].push(fn);
+
+                  data.fieldEventListeners = eventListeners;
+                },
+                off: function(eventName, fn) {
+                  var eventListeners = data.fieldEventListeners[field.name][eventName];
```
tonytlwu, line 1294: `off()` might be executed before `on`, which means `data.fieldEventListeners[field.name][eventName]` will fail due to `data.fieldEventListener` not existing.

I suggest using `_.get(data, ['fieldEventListeners', field.name, eventName])`

js/libs/form.js
```diff
@@ -1276,7 +1277,44 @@ Fliplet().then(function() {
...
+                  }
+
+                  // Remove matching handler
+                  eventListeners.forEach(function(handler, i) {
+                    if (handler === fn) {
+                      eventListeners.splice(i, 1);
+                    }
+                  });
+
+                  data.fieldEventListeners[field.name][eventName] = eventListeners;
```
tonytlwu, line 1314: Can you confirm if this line is necessary? You ran `eventListeners = [];` above and didn't reassign the empty array back to the event listeners -- because it's an object reference. By the same logic, this line wouldn't be needed either.

tonytlwu (changes requested):

js/libs/form.js
armine-fliplet: I forgot to reassign the value in this scope.

js/libs/form.js
tonytlwu: For consistency. Then we can update documentation so that `$instance` (with a `$`) is always a Vue instance object.

js/libs/form.js
```diff
@@ -1072,7 +1072,10 @@ Fliplet().then(function() {
             name: data.name,
             instance: $form,
             data: function() {
```
tonytlwu, line 1074:
```suggestion
            $instance: $form,
            data: function() {
```

js/libs/form.js
```diff
@@ -1072,7 +1072,10 @@ Fliplet().then(function() {
...
-              return data;
+              return {
+                data,
+                $instance: $form
```
tonytlwu, line 1077: Wrong line. See above.

```suggestion
```

tonytlwu (changes requested):

tonytlwu (approved):

PR #498, opened 2023-05-04
ID-3040 [FIX] Restore predefined range field on clear for date range field

armine-fliplet:
… button


tonytlwu (changes requested):
Also, the PR title is truncated, so you'll want to watch out for that and fix it too

tonytlwu (approved):

PR #501, opened 2023-05-19
ID-3067 [FIX] Fixes an issue with incorrect date data after resetting form

armine-fliplet:


js/components/dateRange.js
```diff
@@ -234,21 +216,20 @@ Fliplet.FormBuilder.field('dateRange', {
...
-            end: this.formatDate()
-          };
+          return this.value;
       }
     },
     formatDate: function(date) {
-      return typeof date !== 'undefined' && moment(date).isValid() ? moment(date).locale('en').format('YYYY-MM-DD') : moment(date).locale('en').format('YYYY-MM-DD');
+      return typeof date !== 'undefined' && moment(date).isValid()
+        ? moment(date).locale('en').format('YYYY-MM-DD')
+        : moment(date).locale('en').format('YYYY-MM-DD');
```
tonytlwu, line 225: These two lines are completely the same

js/components/dateRange.js
```diff
@@ -234,21 +216,20 @@ Fliplet.FormBuilder.field('dateRange', {
...
     },
     onBeforeSubmit: function(data) {
       // Empty date fields are validated to null before this hook is called
       if (this.autofill === 'always' && data[this.name] === null) {
         data[this.name] = this.defaultSource === 'submission'
           ? { start: '', end: '' }
```
tonytlwu, line 231: If it uses "submission" as the default source, should it not use the date at the point of submission, instead of an empty string?

js/components/dateRange.js
```diff
@@ -152,9 +138,10 @@ Fliplet.FormBuilder.field('dateRange', {
       this.$emit('_input', this.name, val, false, true);
     },
     selectedRange: function(range) {
+      var $vm = this;
```
tonytlwu, line 141: Why doesn't this work?

```suggestion
      this.value = this.getDate(range.value)
```

tonytlwu (changes requested):

tonytlwu (approved):

PR #503, opened 2023-05-23
ID-3071 [FIX] Ensure time range value is not saved in DS after clicking clear button 

armine-fliplet:


js/libs/form.js
```diff
@@ -339,7 +339,9 @@ Fliplet().then(function() {
             } else if (field._type === 'flDateRange' && ['default', 'always'].indexOf(fieldSettings.autofill) > -1) {
               value = fieldSettings.defaultSource === 'submission' ? moment().locale('en').format('YYYY-MM-DD') : $vm.today;
```
tonytlwu, line 340: Is this not a problem as well?

```suggestion
              value = fieldSettings.defaultSource === 'submission' ? moment().locale('en').format('YYYY-MM-DD') : { start: $vm.today, end: $vm.today };
```

PR #504, opened 2023-05-24
ID-3072 [FIX] Autofill with current time when empty form is submitting

armine-fliplet:


js/libs/form.js
```diff
@@ -639,15 +649,15 @@ Fliplet().then(function() {
...
                   value = JSON.stringify({
-                    start: value.start,
-                    end: value.end
+                    start: !value.start && ['default', 'always'].indexOf(field.autofill) > -1 ? $vm.now : value.start,
+                    end: !value.end && ['default', 'always'].indexOf(field.autofill) > -1 ? $vm.now : value.end
                   });
                 }
 
                 if (type === 'flDateRange' && typeof value === 'object') {
                   value = JSON.stringify({
```
tonytlwu, line 658: I see the stringification was added in https://github.com/Fliplet/fliplet-widget-form-builder/pull/493 Why is it necessary to stringify? We have other fields that also generate Object values, but stringification wasn't necessary, e.g. Matrix field.

tonytlwu: @armine-fliplet How is it stored if you didn't stringify it? And what's the consequence? We're usually OK with an object being stored.

armine-fliplet: @tonytlwu It was storing with `{ end: "", start: "" }` format

tonytlwu: @armine-fliplet I see. Because object keys are not enumerated, we're OK if the start and end keys are in the wrong order.

Based on that, can we store it as objects please? Or is this causing other issues?

tonytlwu: @armine-fliplet That's right. There's no need to stringify

tonytlwu (approved):

PR #508, opened 2023-05-30
ID-3067 [FIX] Custom default value inputs are bound incorrectly

armine-fliplet:


tonytlwu (commented):
Can you explain your fix? The code looks like you've only changed the UI design (which doesn't appear to be the reported issue) and I don't see how that fixes the reported functional bug.

tonytlwu (approved):

PR #514, opened 2023-06-07
ID-3084 [FIX] Ensure Multiple line input show correct default value in Preview mode when the Form contains many fields

NikitaJadhavXoriant:
**Product areas affected**

Studio -> Form -> Multiple line field

**What does this PR do?**

Implemented code so that the Multiple line input show correct default value in Preview mode when the Form contains many fields

**Result**

https://github.com/Fliplet/fliplet-widget-form-builder/assets/108272606/8adccdc5-370d-4d96-b1c8-2d1d3bfd1e81

**Checklist**

None

**Testing instructions**

None

**Deployment instructions**

None

**Author concerns**

None


js/libs/form.js
```diff
@@ -153,8 +153,24 @@ Fliplet().then(function() {
       });
     }
 
+    function saveProgress() {
```
tonytlwu, line 156: Can you define this so that we're not repeating the logic in https://github.com/Fliplet/fliplet-widget-form-builder/blob/master/js/libs/form.js#L1015-L1025 ?

tonytlwu (changes requested):

js/libs/form.js
```diff
@@ -1032,17 +1032,7 @@ Fliplet().then(function() {
...
-
-          $vm.fields.forEach(function(field) {
-            if (field.saveProgress !== false && field.enabled) {
-              progress[field.name] = field.value;
-            }
-          });
-
-          localStorage.setItem(progressKey, JSON.stringify(progress));
-        }, saveDelay);
+        this.saveProgressed = debounce(saveProgress(), saveDelay);
```
tonytlwu, line 1035: @NikitaJadhavXoriant Have you confirmed this works? It looks like your syntax is wrong.

```suggestion
        this.saveProgressed = debounce(saveProgress, saveDelay);
```

tonytlwu (changes requested):

js/libs/form.js
NikitaJadhavXoriant: Hi @tonytlwu, its flow is working fine, we checked, still should i go ahead and remove paranthesis?

tonytlwu: @NikitaJadhavXoriant Yes please. See https://lodash.com/docs/4.17.15#debounce that `_.debounce()` takes a function as the first parameter. What you've current got here is not passing a function as the first parameter.

image

You should see that if the first parameter isn't a function, it generates an error.

tonytlwu (changes requested):

tonytlwu (approved):

PR #516, opened 2023-06-09
ID-3272 [FIX] Set correct default value for empty option on reset

armine-fliplet:


tonytlwu (changes requested):
@armine-fliplet Please fix the code conflict

tonytlwu (approved):

PR #528, opened 2023-06-20
ID-3287 [FIX] Fixes an issue where placeholders and options were shown as untranslated

armine-fliplet:


js/components/dateRange.js
```diff
@@ -46,37 +46,34 @@ Fliplet.FormBuilder.field('dateRange', {
...
       today: this.formatDate(),
-      selectedRange: {
-        label: T('widgets.form.dateRange.rangePlaceholder'),
-        value: ''
-      },
+      selectedRange: null,
       predefinedRanges: [
         {
-          label: T('widgets.form.dateRange.predefinedRanges.today'),
+          label: 'Today',
```
tonytlwu, line 52: The labels for predefined ranges need to be translated because these labels are not provided by the user

js/components/dateRange.js
```diff
@@ -36,7 +36,7 @@ Fliplet.FormBuilder.field('dateRange', {
     placeholder: {
       type: String,
       default: function() {
-        return '-- ' + T('widgets.form.dateRange.rangePlaceholder');
+        return '-- Select range';
```
tonytlwu, line 39: Can the Studio user customise this? If not, then we do need continue to use `T()` for this and show it to the user.

tonytlwu (changes requested):

tonytlwu (approved):

PR #529, opened 2023-06-21
ID-3269 [FEAT] Ensure to update matrix field to work with data in different DS columns

NikitaJadhavXoriant:
**Product areas affected**

Studio -> Form -> Matrix 

**What does this PR do?**

Implemented code to update matrix field to work with data in different DS columns

**Result**

https://github.com/Fliplet/fliplet-widget-form-builder/assets/108272606/afea3d8d-eafb-4fde-958f-6fa1fc35dab3

**Checklist**

None

**Testing instructions**

None

**Deployment instructions**

None

**Author concerns**

None


js/libs/form.js
```diff
@@ -114,6 +114,41 @@ Fliplet().then(function() {
       };
     }
 
+    function getMatrixValue(value, field) {
+      var matrixOption = {};
```
tonytlwu, line 118:
```suggestion
      var matrixValue = {};
```

js/libs/form.js
```diff
@@ -114,6 +114,41 @@ Fliplet().then(function() {
...
+
+      if (!value || typeof value === 'object') {
+        return value;
+      }
+
+      if (value.indexOf('[') > -1 || value.indexOf(']') > -1) {
+        _.forEach(value.split(/\r?\n/), function(rawOption) {
+          if (rawOption) {
```
tonytlwu, line 126: Use early exit to reduce the amount of indentations

```suggestion
          if (!rawOption) {
            return;
          }
```

js/libs/form.js
```diff
@@ -114,6 +114,41 @@ Fliplet().then(function() {
...
 
+    function getMatrixValue(value, field) {
+      var matrixOption = {};
+
+      if (!value || typeof value === 'object') {
+        return value;
+      }
+
+      if (value.indexOf('[') > -1 || value.indexOf(']') > -1) {
+        _.forEach(value.split(/\r?\n/), function(rawOption) {
```
tonytlwu, line 125: I can't tell if you meant `rowOption` or `rawOption`. `rowOption` appears to make more sense to me.

```suggestion
        _.forEach(value.split(/\r?\n/), function(rowOption) {
```

js/libs/form.js
```diff
@@ -114,6 +114,41 @@ Fliplet().then(function() {
...
+          if (rawOption) {
+            rawOption = rawOption.trim();
+
```
tonytlwu, line 128: If after trimming the option is an empty string, then ignore it (because it could have been a line of spaces

js/libs/form.js
```diff
@@ -114,6 +114,41 @@ Fliplet().then(function() {
...
+            var regex = /\[(.*)\]/g;
+            var match = rawOption.split(regex).filter(r => r !== '');
+
+            if (match.length > 1) {
+              matrixOption[match[0].trim()] =  match[1].trim();
```
tonytlwu, line 133: Use return to avoid using `else`, and therefore reduce the indentations.

```suggestion
              matrixOption[match[0].trim()] =  match[1].trim();
              return;
```

js/libs/form.js
```diff
@@ -114,6 +114,41 @@ Fliplet().then(function() {
...
+            } else {
+              _.forEach(field.rowOptions, function(row) {
+                if (!_.has(matrixOption, row.label)) {
```
tonytlwu, line 136: Early exit

```suggestion
                if (_.has(matrixOption, row.label)) {
                  return;
                }
```

js/libs/form.js
```diff
@@ -1273,7 +1358,21 @@ Fliplet().then(function() {
 
                     var hasChanged = field.value !== value;
 
-                    field.value = value;
+                    if (field._type === 'flMatrix') {
+                      var option = {};
+
+                      _.some(field.rowOptions, function(row) {
```
tonytlwu, line 1364: You haven't used `return true` anywhere to short circuit the `_.some` loop.

tonytlwu (changes requested):

js/libs/form.js
NikitaJadhavXoriant: @tonytlwu, done please review.

js/libs/form.js
```diff
@@ -1278,7 +1377,24 @@ Fliplet().then(function() {
...
-                    field.value = value;
+                    if (field._type === 'flMatrix') {
+                      var options = {};
+
+                      _.some(field.rowOptions, function(row) {
+                        return _.some(_.keys(value), function(key) {
+                          if (row.label === key) {
+                            options[key] = value[key];
+
+                            return;
```
tonytlwu, line 1388:
```suggestion
                            return true;
```

tonytlwu (changes requested):

tonytlwu (approved):

PR #531, opened 2023-06-26
ID-3294 [FIX] Incorrect data was saved from countdown timer

armine-fliplet:


js/components/timer.js
```diff
@@ -125,7 +125,7 @@ Fliplet.FormBuilder.field('timer', {
         this.timerInterval = setInterval(function() {
           var totalSeconds = $vm.initialTimerValue - $vm.updateValue();
 
-          if (Math.round(totalSeconds) === 0) {
+          if (totalSeconds < 0) {
```
tonytlwu, line 128:
```suggestion
          if (totalSeconds <= 0) {
```

tonytlwu (approved):

PR #535, opened 2023-07-06
ID-3293 [FIX] Timer shows error message even if it's running

armine-fliplet:


js/components/timer.js
```diff
@@ -62,6 +63,11 @@ Fliplet.FormBuilder.field('timer', {
     Fliplet.FormBuilder.off('reset', this.reset);
     Fliplet.Hooks.off('beforeFormSubmit', this.beforeFormSubmit);
   },
+  computed: {
+    runningRequired: function() {
```
tonytlwu, line 67:
```suggestion
    hasRequiredError: function() {
```

tonytlwu (changes requested):

PR #539, opened 2023-07-14
ID-3341 [FEAT] Create Typeahead field type

armine-fliplet:


js/libs/form.js
```diff
@@ -1428,58 +1424,76 @@ Fliplet().then(function() {
...
+                    result = Promise.resolve(result);
                   }
 
-                  if (field._type === 'flSelect') {
-                    // remove all invalid options
-                    _.remove(values, function(val) {
-                      return !(_.isObject(val) || _.isNumber(val) || (_.isString(val) && val.trim()));
-                    });
-                  }
+                  return result.then(function(newOptions) {
```
tonytlwu, line 1437:
```suggestion
                  return Promise.resolve(result).then(function(newOptions) {
```

js/libs/form.js
```diff
@@ -1428,58 +1424,76 @@ Fliplet().then(function() {
...
-                  if (typeof values === 'undefined') {
+                  var result = typeof values === 'function' ? values() : values;
+
+                  if (typeof result === 'undefined') {
                     return field.options;
                   }
 
-                  if (!Array.isArray(values)) {
-                    throw new Error('Options must be an array');
+                  if (!(result instanceof Promise)) {
```
tonytlwu, line 1433: See https://github.com/Fliplet/fliplet-api/pull/5788#discussion_r1263848107

```suggestion
```

tonytlwu (changes requested):

js/libs/core.js
```diff
@@ -697,6 +701,54 @@ Fliplet.FormBuilder = (function() {
...
+            connection.getIndex(column).then(function onSuccess(values) {
+              $vm.options = _.compact(values, function(value) {
+                if (value !== '') {
+                  return value.trim();
+                }
+              }).map(function(rawOption) {
+                if (rawOption) {
+                  rawOption = rawOption.trim();
+
+                  var regex = /<.*>$/g;
```
tonytlwu, line 724: If the data comes from the data source, why would it contain `<*>`?

js/libs/form.js
```diff
@@ -227,24 +227,9 @@ Fliplet().then(function() {
       });
     }
 
-    function saveProgress() {
```
tonytlwu, line 230: This was only just recently added to fix an issue. See PR: https://github.com/Fliplet/fliplet-widget-form-builder/pull/514

Can you justify why you're moving this again, and have you checked it doesn't reintroduce the bug?

js/components/typeahead.js
```diff
@@ -0,0 +1,160 @@
...
+    maxItems: {
+      type: Number,
+      default: null
+    },
+    options: {
+      type: Array,
+      default: function() {
+        return [
+          {
+            id: '1',
```
tonytlwu, line 31: Why is this given a different ID? I'm comparing this with other multi-option fields like https://github.com/Fliplet/fliplet-widget-form-builder/blob/27f5e2bc6810a126e490d06515807c8bdf66c79b/js/components/checkbox.js#L21

js/libs/form.js
```diff
@@ -1455,58 +1451,72 @@ Fliplet().then(function() {
...
+                      if (typeof option === 'object') {
+                        if (typeof option.value !== 'undefined') {
+                          option.id = option.value;
+                        }
 
-                  if (!_.isEmpty(field.value)) {
-                    switch (field._type) {
-                      case 'flCheckbox':
-                        var selectedValues = _.difference(field.value, values);
+                        if (field._type === 'flTypeahead' && typeof option.id === 'undefined') {
```
tonytlwu, line 1480: What are you trying to do with all these logic that's treating the typeahead field differently?

templates/configurations/typeahead.interface.hbs
```diff
@@ -0,0 +1,82 @@
...
+      <span class="hide-field">Allow user to add their options</span>
+    </label>
+  </div>
+</div>
+
+<div v-show="optionsType === 'dataSource'" class="form-group clearfix">
+  <div id="data-provider"></div>
+  <label>Data field name</label>
+  <select class="form-control" :value="column" v-model="column" @change="_getDataSourceColumnValues">
+    <option value="">--- Select a data field</option>
```
tonytlwu, line 33:
```suggestion
    <option value="">-- Select a data field</option>
```

templates/configurations/typeahead.interface.hbs
```diff
@@ -0,0 +1,82 @@
...
+  <input class="form-control" type="text" v-model.number="maxItems" />
+  <div class="secondary-label">Leave it blank to allow any amount of items or enter a number</div>
+</div>
+
+<div class="form-group">
+  <hr/>
+</div>
+
+<div class="form-group">
+  <label>Default value type</label>
```
tonytlwu, line 53: I see all of this being replicated from the `form.interface.hbs` file. Can you set up another .hbs file for this to avoid replicating the HTML markup? It's quite a complex set of markup to be repeated in different files.

tonytlwu (changes requested):

js/libs/core.js
```diff
@@ -697,6 +701,54 @@ Fliplet.FormBuilder = (function() {
...
+
+        Fliplet.Cache.get({
+          key: id + '-' + column,
+          expire: 60
+        }, function getColumnValues() {
+          // If there's no cache, return new values, i.e.
+          Fliplet.DataSources.connect(id).then(function(connection) {
+            connection.getIndex(column).then(function onSuccess(values) {
+              $vm.options = _.compact(values, function(value) {
+                if (value !== '') {
```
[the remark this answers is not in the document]
armine-fliplet: I think not. I have tried to set numbers in DS and seems it's returned as a string.

js/libs/core.js
```diff
@@ -697,6 +701,54 @@ Fliplet.FormBuilder = (function() {
...
+        var column = this.column;
+
+        Fliplet.Cache.get({
+          key: id + '-' + column,
+          expire: 60
+        }, function getColumnValues() {
+          // If there's no cache, return new values, i.e.
+          Fliplet.DataSources.connect(id).then(function(connection) {
+            connection.getIndex(column).then(function onSuccess(values) {
+              $vm.options = _.compact(values, function(value) {
```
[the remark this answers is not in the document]
armine-fliplet: We have such code [here](https://github.com/Fliplet/fliplet-widget-form-builder/blob/master/js/libs/core.js#L900). 
_.compact() supports a second parameter which are values to be concatenated

js/libs/core.js
armine-fliplet: I had considered it would contain

js/libs/core.js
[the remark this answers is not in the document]
tonytlwu: @armine-fliplet You'll find that it's actually using the function as a second parameter to `_.map()`, not `_.compact()`.

js/libs/core.js
tonytlwu: If one of the unique value from data source entries is `Bar <Foo>`, then `Bar <Foo>` is the value. There's no need to parse it further.

js/libs/form.js
armine-fliplet: @tonytlwu I was having issue from selectize.js. It's required to have `id` in the options

tonytlwu: I see. Yes I think it does need to have IDs. You can use the same ID as the label (unless the user has provided a custom ID).

js/libs/core.js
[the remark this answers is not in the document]
tonytlwu: If you're using the data source manager, it's possible that it's all translated to string. If you really want to be sure, use the database to set various different values with different types: object, number, boolean etc.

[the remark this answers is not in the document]
armine-fliplet: So, I should convert the not string values to string or just assign the value as `null` in case of that?

js/libs/core.js
```diff
@@ -697,32 +694,22 @@ Fliplet.FormBuilder = (function() {
...
-                  }
-
-                  return option;
+                if (typeof option === 'object' || Array.isArray(option)) {
+                  return JSON.stringify(option);
                 }
-              });
+
+                return option.toString().trim();
+              }).map(function(rowOption) {
```
tonytlwu, line 707: You should reduce this to use just a single `_.map()` instead of looping through the content twice.

tonytlwu (changes requested):

js/libs/core.js
```diff
@@ -700,14 +700,12 @@ Fliplet.FormBuilder = (function() {
...
-                  return JSON.stringify(option);
+                  option = JSON.stringify(option);
                 }
 
-                return option.toString().trim();
-              }).map(function(rowOption) {
                 return {
-                  id: rowOption,
-                  label: rowOption
+                  id: option.toString().trim(),
```
tonytlwu, line 707:
```suggestion
                  id: (typeof option === 'string' ? option : option.toString()).trim(),
```

tonytlwu (changes requested):
Please fix the merge conflict as well

tonytlwu (approved):

PR #545, opened 2023-07-20
ID-3383 [FEAT] Create Geolocation field type

armine-fliplet:


js/components/geolocation.js
```diff
@@ -0,0 +1,175 @@
...
+    setValue: function(result) {
+      return `${result.coords.latitude}, ${result.coords.longitude}`;
+    },
+    getLocation: function() {
+      var $vm = this;
+
+      $vm.buttonClicked = true;
+      $vm.isLoading = true;
+
+      var location = Fliplet.Navigator.location({
```
tonytlwu, line 62: I would turn

```
Fliplet.Navigator.location({
        maximumAge: 0,
        timeout: 30000,
        enableHighAccuracy: true
      })
```

into a `getDeviceLocation` function so that the API options can be centralised in one place

js/components/geolocation.js
```diff
@@ -0,0 +1,175 @@
...
+        $vm.firstTimeSaved = true;
+        $vm.isLoading = false;
+
+        setTimeout(function() {
+          $vm.buttonClicked = false;
+        }, 3000);
+      }).catch(function(err) {
+        $vm.isLoading = false;
+        $vm.buttonClicked = false;
+        $vm.openToastMessage(err.code);
```
tonytlwu, line 80:
```suggestion
        $vm.openToastMessage(err);
```

js/components/geolocation.js
```diff
@@ -0,0 +1,175 @@
...
+      });
+    },
+    getErrorMessage: function(code) {
```
tonytlwu, line 109:
```suggestion
    getErrorMessage: function(error) {
```

js/components/geolocation.js
```diff
@@ -0,0 +1,175 @@
...
+      switch (code) {
```
tonytlwu, line 110:
```suggestion
      error = error || {};
      
      switch (error.code) {
```

js/components/geolocation.js
```diff
@@ -0,0 +1,175 @@
...
+          break;
+      }
+    },
+    openToastMessage: function(code) {
+      var supportsSettings = Fliplet.Navigator.supportsAppSettings();
+
+      Fliplet.UI.Toast({
+        type: 'minimal',
+        message: this.getErrorMessage(code),
+        actions: (code === 1 || code === 'inaccurateCoords') && supportsSettings ? [{
```
tonytlwu, line 131: There should be a linebreak before the `?` operator instead of after it, e.g.

```
actions: condition
  ? value1
  : value 2
```

js/components/geolocation.js
```diff
@@ -0,0 +1,175 @@
...
+          label: 'SETTINGS',
```
tonytlwu, line 132: This is a text label so it should be localised too

js/components/geolocation.js
```diff
@@ -0,0 +1,175 @@
...
+        }] : null,
+        duration: false, // Ensures the toast message doesn't auto-dismiss
+        tapToDismiss: false // Ensures the toast message is only dismissed through the action button
+      });
+    },
+    onBeforeSubmit: function() {
+      if (this.isLoading) {
+        Fliplet.UI.Toast({
+          type: 'regular',
+          message: 'Loading location ...',
```
tonytlwu, line 145: Should be localised

js/components/geolocation.js
```diff
@@ -0,0 +1,175 @@
...
+          actions: [
+            {
+              label: 'Ok',
```
tonytlwu, line 148: Should be localised

js/components/geolocation.js
```diff
@@ -0,0 +1,175 @@
...
+          ]
+        });
+
+        return Promise.reject('');
+      } else if (this.preciseLocationRequired && this.accuracy > 100) {
+        this.openToastMessage('inaccurateCoords');
+
+        return Promise.reject('');
+      } else if (!this.value && !this.required) {
+        this.value = 'No data provided';
```
tonytlwu, line 162: If there's no value, just send `null` and not the "No data provided" string. If you're following the spec, please point out where so we correct the spec.

js/components/geolocation.js
```diff
@@ -0,0 +1,175 @@
...
+    preciseLocationRequired: {
+      type: Boolean,
+      default: false
+    }
+  },
+  data: function() {
+    return {
+      firstTimeSaved: false,
+      isLoading: false,
+      buttonClicked: false,
```
tonytlwu, line 24:
```suggestion
      showFeedback: false,
```

js/components/geolocation.js
```diff
@@ -0,0 +1,175 @@
...
+
+      location.then(function(result) {
+        $vm.accuracy = result.coords.accuracy;
+        $vm.value = $vm.setValue(result);
+        $vm.firstTimeSaved = true;
+        $vm.isLoading = false;
+
+        setTimeout(function() {
+          $vm.buttonClicked = false;
+        }, 3000);
```
tonytlwu, line 76: Have you considered what happens when the user clicks on the get/update location button multiple times?

tonytlwu (changes requested):

js/components/geolocation.js
armine-fliplet: @tonytlwu see https://www.awesomescreenshot.com/video/19437600

tonytlwu (changes requested):

js/components/geolocation.js
```diff
@@ -0,0 +1,186 @@
...
+    Fliplet.FormBuilder.on('reset', this.onReset);
+    Fliplet.Hooks.on('beforeFormSubmit', this.onBeforeSubmit);
+  },
+  destroyed: function() {
+    Fliplet.FormBuilder.off('reset', this.onReset);
+    Fliplet.Hooks.off('beforeFormSubmit', this.onBeforeSubmit);
+  },
+  methods: {
+    setValue: function(result) {
+      return `${result.coords.latitude}, ${result.coords.longitude}`;
```
tonytlwu, line 55: Although we do allow spaces when we read data from a data source, the value format should not have a space when it's generated by the form. Please check this is the case everywhere.

```suggestion
      return `${result.coords.latitude},${result.coords.longitude}`;
```

tonytlwu, line 55: I think it's best to check for `!result` or `!result.coords` just in case and return `null`.

tonytlwu, line 55: This should just set the `this.value` variable instead of returning a value. That's what the function is called, "set value".

tonytlwu (changes requested):

js/components/geolocation.js
```diff
@@ -52,7 +52,11 @@ Fliplet.FormBuilder.field('geolocation', {
   },
   methods: {
     setValue: function(result) {
-      return `${result.coords.latitude}, ${result.coords.longitude}`;
+      if (!result || !result.coords) {
+        this.value = null;
+      } else {
```
tonytlwu, line 57:
```suggestion
        return;
      }
```

tonytlwu (changes requested):

PR #547, opened 2023-08-02
ID-3209 [FIX] Ensure Matrix field does not creates extra columns in DS if the form has extra field, when user submits it

NikitaJadhavXoriant:
**Product areas affected**

Studio -> Form -> Matrix

**What does this PR do?**

Implemented code, so when user submit Form with matrix and other field, it should not create extra Matrix column.

**JIRA ticket**

[click here](https://weboo.atlassian.net/browse/ID-3375)

**Result**

https://github.com/Fliplet/fliplet-widget-form-builder/assets/108272606/ba0742c0-73d8-4436-aea5-f28d994b36f0

**Checklist**

None

**Testing instructions**

None

**Deployment instructions**

None

**Author concerns**

None

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -1040,6 +1040,16 @@ Fliplet().then(function() {
                     append: false
                   };
                 }
+
+                if (field._type === 'flMatrix') {
+                  formData._flSchema.excludedFields = [];
+
+                  _.forEach(field.rowOptions, function(row) {
+                    var val = row.id ? row.id : row.label;
+
+                    formData._flSchema.excludedFields.push(`${field.name} [${val}]`);
+                  });
+                }
               });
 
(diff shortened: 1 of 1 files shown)


js/libs/form.js
```diff
@@ -1040,6 +1040,16 @@ Fliplet().then(function() {
                     append: false
                   };
                 }
+
+                if (field._type === 'flMatrix') {
+                  formData._flSchema.excludedFields = [];
+
+                  _.forEach(field.rowOptions, function(row) {
+                    var val = row.id ? row.id : row.label;
```
tonytlwu, line 1048:
```suggestion
                    var val = row.id || row.label;
```

tonytlwu (changes requested):

PR #553, opened 2023-08-08
ID-3451 [FIX] Fixes an issue where some of the form fields have empty settings

armine-fliplet:


templates/configurations/form.interface.hbs
```diff
@@ -175,7 +175,7 @@
 
     {{#unless isTypeahead}}
     <div v-if="_componentsWithPersonalization.includes(_componentName)" class="form-group cleafix">
-      {{> defaultValue}}
+      {{defaultValue}}
```
tonytlwu, line 178: Interesting. Can you explain why this worked? I thought partials are rendered through `{{> }}`? https://handlebarsjs.com/guide/partials.html

armine-fliplet: Honestly, I don't know, the error occurred with `>` this for form interface, but it worked fine for typeahead interface.

tonytlwu: @armine-fliplet Could you investigate and find out why? We don't want unintended consequences so we should know why it works for one context but not for the other (or why `>` is required in one but not the other).

armine-fliplet: @tonytlwu see the ChatGPT answer [here](https://chat.openai.com/share/9d4b5d44-13d5-4725-992a-22f07d99a470)

tonytlwu (approved):

PR #561, opened 2023-08-21
ID-3491 [FIX] Reset default options when switching suggested values type

armine-fliplet:


tonytlwu (approved):

Skip compile: Yes
Includes translation: 
Initiative: Typeahead field
Impact: Low
Affected: N/A - New field (typeahead)
Verify: Typeahead configurations are saved and loaded as expected

PR #565, opened 2023-08-24
ID-3520 [FIX] Place fields on the same level in case of different title sizes

armine-fliplet:
image


tonytlwu (approved):

Skip compile: 
Includes translation: 
Impact: High
Affected: All form fields
Verify: Form fields are rendered and laid out as expected

PR #566, opened 2023-08-24
ID-3521 [FIX] Fields have different sizes when sharing and not sharing the row with the same width

armine-fliplet:
One of the fields with the same size has `own` property.

image

### DIFF
--- a/css/form.css
+++ b/css/form.css
@@ -8,7 +8,7 @@
 }
 
 .w-1_4.own {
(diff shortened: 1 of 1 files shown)


tonytlwu (approved):

tonytlwu:

Skip compile: 
Includes translation: 
Impact: Low
Affected: All form fields
Verify: Fields are laid out as expected

PR #567, opened 2023-08-29
ID-3521 [FIX] Fields have small differences in size when sharing and not sharing

armine-fliplet:


tonytlwu (approved):

Skip compile:
Includes translation:
Impact: Low
Affected: All form fields
Verify: Fields are laid out as expected

PR #569, opened 2023-08-30
ID-3528 [FIX] Fixes undefined columns error in getDataSourceColumns method

armine-fliplet:


tonytlwu (approved):

Skip compile:
Includes translation:
Initiative: Typeahead
Impact: Low
Affected: Typeahead fields
Verify: Typeahead fields are using correct placeholder text

PR #570, opened 2023-08-30
ID-3540 [FIX] Changed flexible width prop name to not have issue with image field

armine-fliplet:


tonytlwu (approved):

Skip compile: 
Includes translation: 
Impact: High
Affected: All fields
Verify: All fields are rendered as expected

PR #577, opened 2023-09-08
ID-3571 [FIX] Fixes form components' width to match container size

armine-fliplet:
### Result:

https://github.com/Fliplet/fliplet-widget-form-builder/assets/128052346/ad4b87ee-a089-4eb8-9003-80b4f439e365

### DIFF
--- a/css/form.css
+++ b/css/form.css
@@ -6,7 +6,6 @@
   display: flex;
   flex-direction: column;
   margin: 0;
-  margin-bottom: 15px;
 }
 
 .col-xs-12 {
@@ -31,7 +30,7 @@
 }
 
 .w-1_4.own {
-  width: 24%;
+  width: calc(25% - 9px);
   grid-column-end: span 4;
 }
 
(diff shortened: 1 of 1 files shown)


css/form.css
```diff
@@ -1524,6 +1524,10 @@ form hr {
     display: flex !important;
     align-items: center;
   }
+
+  .scanner.btn.btn-primary {
+    margin-bottom: 8px;
```
tonytlwu, line 1529: Why is this here? Doesn't seem related to custom field widths.

armine-fliplet: Removed!

css/form.css
```diff
@@ -6,6 +6,7 @@
   display: flex;
   flex-direction: column;
   margin: 0;
+  margin-bottom: 15px;
```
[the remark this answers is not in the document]
armine-fliplet: I have changed the grid gap from `px` to `%` and it's showing too small size for the row gap. That's why I set the gap only for the columns and added margin-bottom for all fields.

tonytlwu (approved):

Skip compile: 
Includes translation: 
Impact: High
Affected: All form instances
Verify: Forms are laid out on the page as expected

PR #578, opened 2023-09-08
ID-3568 [FIX] Typeahead newly added values in DS disappear and can’t be selected

armine-fliplet:
### Result:

https://github.com/Fliplet/fliplet-widget-form-builder/assets/128052346/0bf3e9b6-e8d7-431d-a187-7fcc9b1ecc6c

### DIFF
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -574,6 +574,7 @@ Fliplet().then(function() {
         }
       },
       methods: {
+        getDataSourceColumnValues,
         saveProgress: function() {
           var progress = {};
 
--- a/js/libs/form.js
+++ b/js/libs/form.js
@@ -64,6 +64,44 @@ function addThumbnailToCanvas(imageURI, indexCanvas, self, isFileCanvas) {
   });
 }
 
+function getDataSourceColumnValues(field) {
+  var id = field.dataSourceId;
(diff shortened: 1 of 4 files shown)


js/libs/form.js
```diff
@@ -64,6 +64,44 @@ function addThumbnailToCanvas(imageURI, indexCanvas, self, isFileCanvas) {
   });
 }
 
+function getDataSourceColumnValues(field) {
```
tonytlwu, line 67: I see the same function declared in https://github.com/Fliplet/fliplet-widget-form-builder/blob/f870c19befb4c95f4d0ac4b41f51f2c6e0d07bd3/js/libs/core.js#L14

How can we avoid redeclaring what's essentially the same logic?

tonytlwu (changes requested):

js/libs/form.js
armine-fliplet: Yes, they have the same logic. Where I can declare this method as I can use it in both form.js and core.js files?

tonytlwu (approved):

Skip compile: 
Includes translation: 
Initiative: 
Impact: High
Affected: Typeahead fields
Verify: Typeahead fields are working as expected

PR #579, opened 2023-09-11
ID-3570 [FIX] Ensure to get correct default value from user profile data

armine-fliplet:


squallstar (approved):
@inna-bieshulia this is ready for testing.

- Risk: low
- Affected: form fields loading data
- Verify: form fields load data as expected

PR #581, opened 2023-09-22
ID-3604 [FIX] Ensure form error message has full width

armine-fliplet:
### Result

image


tonytlwu (approved):

Skip compile: 
Includes translation: 
Impact: Medium
Affected: All form instances
Verify: UI layout is correct for all form UI and on all device types

PR #584, opened 2023-09-29
ID-3480 [FIX] Add a separate column for location accuracy

armine-fliplet:
### Result

Geolocation

### DIFF
--- a/js/components/geolocation.js
+++ b/js/components/geolocation.js
@@ -3,7 +3,8 @@ Fliplet.FormBuilder.field('geolocation', {
   category: 'Advanced',
   props: {
     value: {
-      type: String
+      type: Array,
+      default: null
     },
     description: {
       type: String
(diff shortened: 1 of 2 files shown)


js/components/geolocation.js
```diff
@@ -3,7 +3,8 @@ Fliplet.FormBuilder.field('geolocation', {
   category: 'Advanced',
   props: {
     value: {
-      type: String
+      type: Array,
```
tonytlwu, line 6: Why is it that we need to change the variable type for `value` and can't just keep the `accuracy`?

armine-fliplet: I tried to create an `accuracy` property, but I needed to update the value in `form.js` file. I developed an emit method, but somehow it didn't work with `_input` event. If this solution doesn't please you, I can try finding another solution for `accuracy` prop.

tonytlwu: 👍🏻

PR #585, opened 2023-10-11
ID-3659 [Fix] Updates selector for styling field row containers

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-3659

https://github.com/Fliplet/fliplet-widget-form-builder/commit/837b5c3f6ad46963a03b71e9298d9bcd4bfe6b14 implemented `display:flex` based on the `.form-group.row` selector. However, there are custom CSS in apps that use `[data-field="fieldName"]` which no longer works because `.form-group.row` has a higher specificity.

This PR fixes the problem by introducing a new attribute selector that has the same level of specificity as `[data-field="fieldName"]` so that custom CSS will work as expected.

This also improves the system by avoiding styling generic Bootstrap classes such as `.form-group.row` or `.col-xs-12` as we don't want our component to affect how Bootstrap works in the wider context.


armine-fliplet:
@tonytlwu this works for me.
Only [this line](https://github.com/Fliplet/fliplet-widget-form-builder/blob/66cda39283c2bfed540169e1db1ccc534efdeeb1/templates/components/password.build.hbs#L60C53-L60C53) also needs the `data-form-field` attribute.

About custom styling, what type of changes could the app creator make? I mean if he changes the width of field with the `[data-field="fieldName"]` selector, it will not work, is it ok?

simonchapman1986 (approved):
LGTM 🚀

tonytlwu:

Skip compile: 
Includes translation: 
Impact: High
Affected: Form field layout
Verify: Form fields and widths are laid out as expected

PR #587, opened 2023-10-18
ID-3658 [FIX] Fixes getting typeahead options from DS and setting correct value from LFD

armine-fliplet:
#### Result 

https://github.com/Fliplet/fliplet-widget-form-builder/assets/128052346/4a982de3-f15e-4968-93e2-79b759ad82e5


tonytlwu (commented):
I see a lot of the form logic being removed, and moved to only work in the form builder context. Can you explain how this solution works, i.e. what was the problem and how does your solution fix the problem (without introducing new problems)?

js/libs/builder.js
```diff
@@ -323,6 +323,44 @@ Fliplet().then(function() {
...
+        if (!id || !column) {
+          return;
+        }
+
+        Fliplet.Cache.get({
+          key: id + '-' + column,
+          expire: 60
+        }, function getColumnValues() {
+          // If there's no cache, return new values, i.e.
+          Fliplet.DataSources.connect(id).then(function(connection) {
```
tonytlwu, line 339: This looks like a problem from the beginning, before your change, because the Promise request isn't being returned.

```suggestion
          return Fliplet.DataSources.connect(id).then(function(connection) {
```

js/libs/builder.js
```diff
@@ -323,6 +323,44 @@ Fliplet().then(function() {
...
+            connection.getIndex(column).then(function onSuccess(values) {
```
tonytlwu, line 340: Doesn't the request need to be returned for the promise to work?

Also, the `.then()` can be chained instead of nested.

```suggestion
            return connection.getIndex(column);
          }).then(function onSuccess(values) {
```

tonytlwu (changes requested):

js/libs/builder.js
```diff
@@ -323,6 +323,44 @@ Fliplet().then(function() {
           Fliplet.Studio.emit('widget-mode', 'wide');
         }
       },
+      getDataSourceColumnValues: function(field) {
```
simonchapman1986, line 326: one thing that catches my attention, not only the promises that @tonytlwu mentioned, but also "callback hell". Might be worth breaking it up a bit to make it a little cleaner with awaits.

thoughts @tonytlwu @dwfliplet ?

```
async function getDataSourceColumnValues(field) {
  const id = field.dataSourceId;
  const column = field.column;

  if (!id || !column) {
    return;
  }

  try {
    const cachedValues = await Fliplet.Cache.get({
      key: `${id}-${column}`,
      expire: 60
    });

    if (cachedValues) {
      field.options = cachedValues;
    } else {
      const connection = await Fliplet.DataSources.connect(id);
      const values = await connection.getIndex(column);
      field.options = values.map(option => ({
        id: option ? option.toString().trim() : '',
        label: option ? option.toString().trim() : ''
      }));
      Fliplet.Cache.set({ key: `${id}-${column}`, value: field.options });
    }
  } catch (error) {
    // Handle errors, e.g., log or provide user feedback
    console.error('Error in getDataSourceColumnValues:', error);
  }
}
```

js/libs/core.js
```diff
@@ -83,7 +83,7 @@ Fliplet.FormBuilder = (function() {
        *
        * @return {void} - this method does not return anything but updates the value directly on the target field.
        */
-      component.methods.setValueFromDefaultSettings = function(data) {
+      component.methods.setValueFromDefaultSettings = async function(data) {
```
armine-fliplet, line 86: @tonytlwu This is for getting the default value before getting the value from LFD.
While I was testing, I noticed that when the field default value is from user data and I'm trying to edit LFD entry, the default value is set later than the entry value.

js/libs/form.js
```diff
@@ -72,35 +72,32 @@ function getDataSourceColumnValues(field) {
     return;
```
tonytlwu, line 72: This doesn't return a Promise

js/components/typeahead.js
```diff
@@ -73,32 +73,6 @@ Fliplet.FormBuilder.field('typeahead', {
     return rules;
   },
   created: function() {
-    var $vm = this;
```
tonytlwu, line 76: Why is all this deleted?

js/libs/form.js
```diff
@@ -491,7 +501,12 @@ Fliplet().then(function() {
 
                 break;
                 // There is no validation and value assignment for checkbox and radio options as there is no access to the options. This is implemented in the checkbox and radio components respectively.
-
+              case 'flTypeahead':
+                getDataSourceColumnValues(field).then(function(result) {
```
tonytlwu, line 505: Why are we no longer checking for `field.optionsType === 'dataSource'`?

js/libs/form.js
```diff
@@ -519,15 +534,18 @@ Fliplet().then(function() {
...
-            getDataSourceColumnValues(field);
-          }
-
           if (progress && !isEditMode) {
             var savedValue = progress[field.name];
 
             if (typeof savedValue !== 'undefined') {
-              field.value = savedValue;
+              if (field._type === 'flTypeahead') {
+                getDataSourceColumnValues(field).then(function(result) {
```
tonytlwu, line 542: Same as above, why are not checking for `field.optionsType === 'dataSource'?

js/libs/form.js
```diff
@@ -97,10 +104,13 @@ function getDataSourceColumnValues(field) {
         return field.options;
       });
     });
-  })
-    .then(function(result) {
-      field.options = result;
-    });
+  }).then(function(result) {
+    field.options = result;
```
tonytlwu, line 108: You've used `getDataSourceColumnValues(field).then((result) => { field.options = result; })` many times in other places in this file, but this line is already here. One of them seems unnecessary (I think this one is).

js/libs/form.js
```diff
@@ -1495,6 +1515,11 @@ Fliplet().then(function() {
                       });
 
                       field.value = options;
+                    } else if (field._type === 'flTypeahead') {
+                      getDataSourceColumnValues(field).then(function(result) {
```
tonytlwu, line 1519:
```js
                      getDataSourceColumnValues(field).then(function(result) {
                        field.options = result;
                        field.value = value;
                      });
```

You've used the pattern above several times. I suggest creating a `setTypeaheadFieldValue(field, value)` function to do this

tonytlwu (changes requested):

js/components/typeahead.js
armine-fliplet: this is an unnecessary part of code. The field options are not rendered here, which could be a reason for losing the value.

tonytlwu (approved):

Skip compile: 
Includes translation: 
Initiative: Typeahead field
Impact: High
Affected: Studio and app users using the Typeahead field
Verify: The field functions as expected, especially when loading data from a data source entry, setting default values and resetting forms

PR #592, opened 2023-11-16
ID-3771 [FIx] Updates default placeholder value

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-3771


simonchapman1986 (approved):
Risk: low
Affected: typeahead
Verify:
- default value no longer has the excessive space between `start typing` and the `...`

PR #593, opened 2023-11-20
ID-3480 [FIX] Prevent form submitting when precise location is required but the device has this feature disabled

armine-fliplet:


simonchapman1986 (approved):
Risk: low
Affected: form submission w/ location
Verify:
- prevent form submission when location required and location services disabled

PR #603, opened 2024-01-30
ID-3977 [FIX] Move rich text '...' to the right corner on mobile and tablet

armine-fliplet:
### Result 
image
image
image


simonchapman1986 (approved):
Risk: low
Affected: rich text editor
Verify: position of text is in right corner on mobile/tablet

PR #604, opened 2024-01-31
ID-3976 [FIX] Update field value while undoing and redoing

armine-fliplet:


simonchapman1986 (approved):
Risk: low
Affected: RichText
Verify: undo/redo operates as expected

PR #608, opened 2024-02-06
ID-3993 [FIX] Ensure typeahead not triggers infinite loop

armine-fliplet:
When we are trying to assign a field value from LFD that contains more items than the field max items value, the field triggers an infinite loop.

**What does this PR do?**

Splits the field value with the max items count if it is provided.

### Result 
https://www.loom.com/share/dc7e30b48c0347579afdec4ddd78eed2?sid=9d11d0a6-355e-4bed-9377-3a50686dd9b4

### DIFF
--- a/js/components/typeahead.js
+++ b/js/components/typeahead.js
@@ -118,6 +118,10 @@ Fliplet.FormBuilder.field('typeahead', {
   },
   watch: {
     value: function(val) {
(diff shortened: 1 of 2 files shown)


simonchapman1986 (approved):
Risk: low
Affected: typeahead 
Verify: infinite loop is not triggered

PR #609, opened 2024-02-16
ID-4009 [FIX] Ensure to not load user data in the form in add mode

armine-fliplet:


simonchapman1986 (approved):
Risk: medium
Affected: form builder load data in add mode
Verify: form data is present in add mode

PR #621, opened 2024-03-12
ID-4120 [FIX] Ensure to not have scroll bar when using TinyMCE 4+ version

armine-fliplet:


simonchapman1986 (approved):
Risk: low
Affected: tinymce 4+
Verify: ensure to not have scroll bar

PR #628, opened 2024-03-18
ID-3976 [FIX] Update field value while undoing and redoing (#604)

armine-fliplet:
https://weboo.atlassian.net/browse/ID-3976


simonchapman1986 (approved):
LGTM 🚀

PR #642, opened 2024-04-03
ID-4123 [FIX] Field height of the text area is too narrow

armine-fliplet:
Ensure to have proper height when row count is 1

### Result
image


dwfliplet (approved):
👍

simonchapman1986 (approved):
Risk: low
Affected: rich text widget
Verify: height of text area is not too narrow

PR #643, opened 2024-04-08
ID-4239 [fix] fixed emptying date and time inputs meant to load data from ds

dwfliplet:
The problem was with condition race - if the field was set to `autofill: 'empty'` and was meant to load data from DS, then the first rendered field (time or date) was emptied after loading DS data. 

I've prevented emptying fields with `autofill: 'empty'` if they are about to load data from DS.

https://weboo.atlassian.net/browse/ID-4239

### DIFF
--- a/js/components/date.js
+++ b/js/components/date.js
@@ -63,7 +63,7 @@ Fliplet.FormBuilder.field('date', {
       this.empty = false;
     }
 
-    if (this.autofill === 'empty') {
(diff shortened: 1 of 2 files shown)


armine-fliplet (approved):
LGTM :+1:

simonchapman1986 (approved):
Risk: low
Affected: form builder
Verify: load data from DS works as expected

PR #641, opened 2024-04-02
Release/ID-4037 Reorder list field

armine-fliplet:
This is a release for the new Reorder list form field

ticket - [ID-4037](https://weboo.atlassian.net/browse/ID-4037)

[ID-4037]: https://weboo.atlassian.net/browse/ID-4037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #645, opened 2024-04-10
Release ID-4239 prevent emptying date and time fields instead of populating them with DS data

dwfliplet:
To be merged after #640 #641 #646


simonchapman1986 (approved):
LGTM 🚀

PR #647, opened 2024-04-10
ID-4144 [FIX] Ensure typeahead list appears in the order we type in

armine-fliplet:
### Result

image
image


dwfliplet (approved):
👌

simonchapman1986 (approved):
Risk: low
Affected: typeahead field
Verify: order is maintained

PR #660, opened 2024-08-28
Feature/ID-4218

armine-fliplet:
### Product areas affected

Form Builder, Address Field

### What does this PR do?

Integrates Address field type and settings

### JIRA ticket

[ID-4218](https://weboo.atlassian.net/browse/ID-4218)

[ID-4218]: https://weboo.atlassian.net/browse/ID-4218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Arpanexe (approved):
LGTM

js/components/address.js
```diff
@@ -0,0 +1,298 @@
...
+            addressData.state = component.long_name;
+          }
+
+          if (types.includes('country')) {
+            addressData.country = component.long_name;
+          }
+
+          if (types.includes('postal_code')) {
+            addressData.postalCode = component.long_name;
+          }
```
dwfliplet, line 162: @armine-fliplet , is that purposely assigned with the same value?

js/components/address.js
```diff
@@ -0,0 +1,298 @@
...
+      separateFieldsName.forEach(field => {
+        field.value = addressData[field.key];
+      });
+    },
+    updateSelectedFieldsProperty: function(attr, value) {
+      if (attr === 'readonly') {
+        value = !value;
+      }
+
+      var fields = this.$parent.fields;
```
dwfliplet, line 180: @armine-fliplet, please avoid using `var`

js/components/address.js
```diff
@@ -0,0 +1,298 @@
...
+      this.addressField = Fliplet.UI.AddressField(this.$refs.addressField);
+      await this.addressField.getAutocompleteSuggestions(input, countryRestrictions)
+        .then(suggestions => {
+          if (typeof this.value === 'object') {
+            this.addressSuggestions = [];
+            this.suggestionSelected = true;
+          } else {
+            this.addressSuggestions = suggestions;
+          }
+        });
```
dwfliplet, line 215: @armine-fliplet, you should never combine `await` with `then`.

 ```suggestion
      const suggestions = await this.addressField.getAutocompleteSuggestions(input, countryRestrictions);
            if (typeof this.value === 'object') {
            this.addressSuggestions = [];
            this.suggestionSelected = true;
          } else {
            this.addressSuggestions = suggestions;
          }
```

File: js/components/address.js [hunk 4]
```diff
@@ -0,0 +1,298 @@
...
+              if (matchedField) {
+                (function(fieldName, matchedFieldValue) {
+                  Fliplet.FormBuilder.get()
+                    .then(function(form) {
+                      form.field(fieldName).set(matchedFieldValue);
+                    });
+                })(fieldName, matchedField.value);
+              }
+            }
+          }
```
dwfliplet, line 238: @armine-fliplet, few improvements here:
1. use early returns to reduce nesting and cognitive complexity
2. There's no need for the inner self-calling function
```suggestion
            if (!this.selectedFieldOptions[key]) continue;
            
            const matchedField = this.separateFieldsName.find(field => field.key === key);
            const fieldName = this.selectedFieldOptions[key];

            if (!matchedField) continue;
              
            Fliplet.FormBuilder.get()
                .then(function(form) {
                  form.field(fieldName).set(matchedField.value);
                });
          }
```

js/components/address.js
```diff
@@ -0,0 +1,298 @@
...
+          data[this.name] = value;
+        }
+      }
+
+      this.updateSelectedFieldsProperty('value', '');
+    }
+  },
+  watch: {
+    value: function(val) {
+      var $vm = this;
```
dwfliplet, line 271: @armine-fliplet  you never use this variable

js/components/address.js
```diff
@@ -0,0 +1,298 @@
...
+
+      if (typeof val === 'object') {
+        $vm.lastChosenAutocompleteValue = val.label;
+        $vm.suggestionSelected = true;
+        val = val.label;
```
dwfliplet, line 276: @armine-fliplet, there's no need for shadowing `this` with `$vm.` You don't pass it any further. Please use `this` instead.

js/libs/core.js
```diff
@@ -883,6 +888,63 @@ Fliplet.FormBuilder = (function() {
...
+        var fields = $vm.$parent.fields;
+
+        if ($vm.fieldOptions.length < 1) {
+          $vm.fieldOptions = fields.map(function(field) {
+            if (field._type !== 'flButtons' && field._type !== 'flAddress') {
+              return { label: field.label, disabled: false };
+            }
+          }).filter(function(field) {
+            return field;
+          });
```
dwfliplet, line 902: @armine-fliplet, this can be shortened to 

```suggestion
          }).filter(Boolean);
```

js/libs/core.js
```diff
@@ -883,6 +888,63 @@ Fliplet.FormBuilder = (function() {
         component.methods.openFilePicker = component.methods._openFilePicker;
       }
 
+      component.methods._getFieldOptions = function() {
+        var $vm = this;
```
dwfliplet, line 892: @armine-fliplet, again, it's not needed as you never use it elsewhere, where context could be lost

js/libs/core.js
```diff
@@ -883,6 +888,63 @@ Fliplet.FormBuilder = (function() {
...
+        var fields = $vm.$parent.fields;
+
+        if ($vm.fieldOptions.length < 1) {
```
dwfliplet, line 895: @armine-fliplet, that's a pretty mysterious expression. There's only one option, `length === 0`. And since 0 is falsy, you can use it like this:

```suggestion
        if (!this.fieldOptions.length) {
```

js/libs/core.js
```diff
@@ -883,6 +888,63 @@ Fliplet.FormBuilder = (function() {
...
+        $vm.fieldOptions.forEach(option => {
+          option.disabled = assignedValues.has(option.label);
+        });
+      };
+
+      component.methods._initAddressTypeahead = function() {
+        var $vm = this;
+        var countries;
+
+        fetch('https://restcountries.com/v3.1/all').then(function(response) {
```
dwfliplet, line 922: @armine-fliplet, what's that? Why do we create some third-party dependency here? The list of countries doesn't change that often. Especially since we use only part of the info. Please extract the needed data into a separate file, and store it inside the project.

js/components/address.js
armine-fliplet: @dwfliplet Yes)

js/components/address.js
armine-fliplet: @dwfliplet Done

js/components/address.js
armine-fliplet: @dwfliplet Done

js/components/address.js
armine-fliplet: @dwfliplet Done

js/libs/core.js
armine-fliplet: @dwfliplet  Are you suggesting we keep the data static, or should we get it from the API?

dwfliplet: @armine-fliplet to keep data static, but it can be kept in separate file, or even fetched from json file but hosted on our side

Arpanexe (approved):
LGTM

dwfliplet (approved):

PR #679, opened 2024-10-24
Feature/DEV-92

armine-fliplet:
### Product areas affected

Fliplet Form Builder
### What does this PR do?

Creates Map form instance

### JIRA ticket

[DEV-92](https://weboo.atlassian.net/browse/DEV-92)

[DEV-92]: https://weboo.atlassian.net/browse/DEV-92?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #683, opened 2024-10-29
Revert "Feature/DEV-92"

inna-bieshulia:
Reverts Fliplet/fliplet-widget-form-builder#679
due to separating work for custom width


armine-fliplet (approved):
LGTM :+1:

PR #697, opened 2024-12-04
Fix/DEV-230

armine-fliplet:
### Product areas affected

Fliiplet Form Builder, Map field

### What does this PR do?

Ensures Map pin is displayed when loading form with the selected address

### JIRA ticket

[DEV-230](https://weboo.atlassian.net/browse/DEV-230)


js/components/map.js
```diff
@@ -206,8 +202,16 @@ Fliplet.FormBuilder.field('map', {
...
 
           if (this.suggestionSelected && this.lastChosenAutocompleteValue === value.trim()) {
-            this.value = this.mapField.getTotalAddress();
-            this.updateAddressSuggestions();
+            const timeout = this.mapField.getGeocoder() ? 0 : 3000;
+
+            setTimeout(() => {
+              if (this.mapField.getTotalAddress()) {
+                this.value = this.mapField.getTotalAddress();
+              }
```
dwfliplet, line 210: @armine-fliplet is timeout 0 here for intentional move to the end of event loop? If the timeout is not needed here, i'd go rather with something like:

```suggestion
            const assignAddress = () => {
                 const address = this.mapField.getTotalAddress();
                 if (address) this.value = address;
            }
            
            const geocoderExists = !!this.mapField.getGeocoder();

            if (geocoderExists) assingAddress();
            else setTimeout(assignAddress, 3000);
```

js/components/map.js
```diff
@@ -261,24 +265,27 @@ Fliplet.FormBuilder.field('map', {
...
-          this.lastChosenAutocompleteValue = val.address;
-          this.addressSuggestions = [];
-          this.isSelectOnMapClicked = true;
+            this.suggestionSelected = true;
+            this.lastChosenAutocompleteValue = val.address;
+            this.addressSuggestions = [];
+            this.isSelectOnMapClicked = true;
 
-          return;
+            return;
```
dwfliplet, line 287: @armine-fliplet the `return` here is pointless, as it just exits from the setTimeout callback, so it does nothing at the end of function.

Arpanexe (approved):
LGTM

js/components/map.js
armine-fliplet: @dwfliplet It didn't work without SetTimeout, so I just removed` return`.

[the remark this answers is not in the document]
armine-fliplet: @dwfliplet Done

dwfliplet (approved):

PR #710, opened 2025-01-28
Revert "Fix/DEV-379: Restrict Map field to Admins and Impersonating Users Only"

ysolodka:
Reverts Fliplet/fliplet-widget-form-builder#707 due to QA fail.


armine-fliplet (approved):
LGTM :+1:

PR #754, opened 2025-04-10
DEV-423 [Feat] Allow certain file types for narrow picker

zeryabkhan91:
### Product Areas Affected
 - js/interface.template.js
 - templates/configurations/file.interface.hbs
 
### What does this PR do?
This PR allows Studio user to configure allowed file types to access and upload from device's external storage for narrow picker.

### Screenshot
image

PR #758, opened 2025-04-18
DEV-423 [Feat] Allow certain file types for narrow picker

zeryabkhan91:
Same as #754 but base branch has been changed to projects branch


armine-fliplet (approved):
LGTM :+1:

PR #852, opened 2025-11-21
Fix typeahead placeholder bleed when fields are hidden

Komsomol:
Fixes PS-1481 where typeahead fields displayed incorrect placeholders when multiple instances existed and some were hidden via toggle(). Each typeahead after the first would inherit the placeholder from the previous instance (2nd gets 1st's placeholder, 3rd gets 2nd's, etc).

Root cause: Fliplet.UI.Typeahead library reuses/clones DOM elements from hidden instances, causing placeholder attribute to bleed across instances.

Solution: Added ensurePlaceholder() method that forces correct placeholder on DOM select element after initialization and when placeholder prop changes. This prevents cross-instance leakage.

Changes:
- Added ensurePlaceholder() method to explicitly set placeholder attribute on DOM element
- Call ensurePlaceholder() after typeahead initialization
- Added placeholder watcher to reapply on reactivity changes
- Added AI.md with Fliplet widget development guidelines

## Bug Fixes
* Fixed typeahead placeholder handling to prevent unintended leakage between multiple component instances.

## Chores
* Upgraded ESLint configuration to version 9 with updated linting rules and dependencies.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,0 @@
-/vendor/**
-/js/build.templates.js
-/js/interface.templates.js
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -1,0 +1,81 @@
+// Flat ESLint configuration (ESLint v9) for Fliplet Form Builder
+// Converted from legacy .eslintrc style; no parserOptions key used.
+
+const js = require('@eslint/js');
+
+module.exports = [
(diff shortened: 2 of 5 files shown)


galisufyan-327 (dismissed):
LGTM

PR #872, opened 2026-02-12
fix: handle object-type image values in canvas rendering (PS-1420)

Komsomol:
## Summary
- Fixes `ERR_INVALID_URL` when images loaded from data sources are stored as objects (`{url: '...', name: '...'}`) instead of plain URL strings
- Updates both `addThumbnailToCanvas()` in form.js and `onImageClick()` in image.js to extract `.url` from object values
- Affects Events/Exhibitor forms where uploaded images fail to render

## Test plan
- [ ] Upload an image in a form field — should render normally (no regression)
- [ ] Load a form with exhibitor data containing image objects — should render correctly instead of showing broken image
- [ ] Click on a loaded image to preview — should open image preview

### DIFF
--- a/js/components/image.js
+++ b/js/components/image.js
@@ -436,6 +436,11 @@ Fliplet.FormBuilder.field('image', {
             return { url: URL.createObjectURL(img) };
           }
 
+          // Handle image values stored as objects (e.g. { url: '...', name: '...' })
+          if (img && typeof img === 'object') {
+            return { url: img.url || '' };
+          }
+
           return { url: img };
         }),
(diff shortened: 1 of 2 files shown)


farhantariq12b (approved):
LGTM - good defensive guards with proper Blob handling.

Komsomol:
Closing: PS-1420 (object-type image values in canvas) is marked Won't Do in Jira, so this fix is not being pursued. Branch preserved, reopen if that changes. (Backlog cleanup, 24 Jun 2026.)

PR #875, opened 2026-03-09
PS-1633: Ensures Rich text field inherits text and placeholder appearance settings

farhantariq12b:
## JIRA Issue
https://weboo.atlassian.net/browse/PS-1633

## Summary
- Ensures Rich text (TinyMCE) field inherits text input and placeholder styles from appearance settings
- Probes computed styles from a hidden `.form-control` element and injects them into the TinyMCE iframe
- Maintains live synchronisation via MutationObservers to detect CSS changes

## Testing
- Set text input styles (font, color, size etc.) in form appearance → Rich text field now inherits them
- Set placeholder styles → Rich text placeholder now inherits them
- Existing fields continue to work as before

## Risk Assessment
- Risk Level: LOW
- Files Modified: 1
- Type: Bug fix

### DIFF
--- a/js/components/wysiwyg.js
+++ b/js/components/wysiwyg.js
@@ -547,9 +547,11 @@ Fliplet.FormBuilder.field('wysiwyg', {
           setTimeout(updateIframeStyles, 0);
         });
 
-        editor.on('keyup', function() {
-          // If content is still empty, re-allow placeholder to show
-
+        /**
+         * Remove the fl-typing class if the editor content is empty,
+         * allowing the placeholder to reappear.
+         */
+        const clearTypingIfEmpty = () => {
           try {
             const body = editor.getBody && editor.getBody();
 
@@ -569,6 +571,10 @@ Fliplet.FormBuilder.field('wysiwyg', {
           } catch (err) {
             // no-op
           }
+        };
+
+        editor.on('keyup', function() {
+          clearTypingIfEmpty();
         });
 
         editor.on('focus', function() {
@@ -603,26 +609,7 @@ Fliplet.FormBuilder.field('wysiwyg', {
           updateIframeStyles();
 
           // On blur, ensure placeholder can appear again if content is empty
-
-          try {
-            const body = editor.getBody && editor.getBody();
-
-            if (body) {
-              const rawHtml = editor.getContent({ format: 'raw' }) || '';
-              const textOnly = (editor.getContent({ format: 'text' }) || '').replace(/\u00a0/g, ' ').trim();
-              const htmlWithoutEmptyParas = rawHtml
-                .replace(/<p>(?:\s|&nbsp;|<br\s*\/?>)*<\/p>/gi, '')
-                .replace(/<br\s*\/?>/gi, '')
-                .replace(/&nbsp;/gi, '')
-                .trim();
-
-              if (textOnly.length === 0 && htmlWithoutEmptyParas.length === 0) {
-                body.classList.remove('fl-typing');
-              }
-            }
-          } catch (err) {
-            // no-op
-          }
+          clearTypingIfEmpty();
         });
 
         editor.on('change', function() {
--- a/js/components/wysiwyg.js
+++ b/js/components/wysiwyg.js
@@ -83,6 +83,138 @@ Fliplet.FormBuilder.field('wysiwyg', {
 
     this.tinymceId = Fliplet.FormBuilderUtils.kebabCase(this.name) + '-' + $(this.$refs.textarea).parents('[data-form-builder-id]').data('formBuilderId');
 
+    /**
+     * Collect text input and placeholder styles from the host page so the editor
+     * inside the TinyMCE iframe matches the app Appearance settings.
+     *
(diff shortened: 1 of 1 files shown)


zeryabkhan91:
## Code Review

### Summary
Makes the TinyMCE Rich Text field inherit text and placeholder appearance settings. A hidden `.form-control` probe is injected to read computed styles via `getComputedStyle()`; values are applied as `!important` CSS into the TinyMCE iframe via `content_style` at init and a `<style>` tag kept in sync at runtime. Two `MutationObserver`s plus a 6-second polling loop watch for Appearance changes. An `fl-typing` body class hides the placeholder overlay during typing.

---

### Critical Issues (Must Fix Before Merge)

**1. `MutationObserver` on `document.documentElement` with `subtree: true` fires on every class/style change anywhere in the DOM** (`js/components/wysiwyg.js`, `startLiveStyleWatch`)

```js
rootObserver.observe(document.documentElement, {
  attributes: true,
  attributeFilter: ['class', 'style'],
  subtree: true,
  childList: false
});
```

Fires on **any** `class`/`style` attribute change on **any** descendant of `<html>` — every tooltip open/close, modal toggle, hover-state class flip, Vue reactivity update. Each fire schedules `maybeUpdateIframeStyles` → `computeStyleSignature` → `collectEditorStyles`, which **creates an `<input>`, appends it, reads computed style, removes it** — forcing a full style recalculation. In a complex form this thrashes the main thread. Scope the observer to the form root (e.g. `this.$el.closest('form')`) and drop `subtree: true` unless absolutely required.

**2. Two separate `editor.on('focus', ...)` handlers registered**

The existing code already has a focus handler for the `iframeElement`. This PR adds a second one near the bottom that calls `updateIframeStyles()`. Both fire on every focus. Merge the call into the existing handler.

**3. `scheduleStyleRefreshBurst` timers are untracked and fire after editor destruction**

```js
setTimeout(updateIframeStyles, 50);
setTimeout(updateIframeStyles, 300);
```
Neither timer ID is captured or cleared in `stopLiveStyleWatch`. If the editor is destroyed within 300ms of init (rapid navigation), the callback fires against a torn-down editor. Same applies to the `setTimeout(updateIframeStyles, 0)` added in `keydown`. Capture all timer IDs and clear them in the `remove` handler.

---

### Warnings (Should Address)

**4. Significant CSS/DOM duplication between init and runtime paths** — `bodyCss`/`placeholderCss`/`typingHideCss` strings are built twice (once outside `setup`, again inside `updateIframeStyles` as `*Live`). Extract a single `buildCss(styles, hasPlaceholder)` helper.

**5. `collectEditorStyles()` does DOM mutations on every call** and is invoked from hot paths (`change`, `undo`, `redo`, `blur`, `focus`, `keydown`, every MutationObserver tick, polling loop). Even with the signature short-circuit, the probe-inject cost fires every time. Cache the probe element for the editor's lifetime and remove only in the `remove` handler.

**6. 11 empty `catch (e) { // no-op }` blocks silently swallow all diagnostics.** If any path throws, the feature stops working with zero signal. At minimum, `console.warn('[fl-rt-style]', e)` with a guard to avoid spam.

**7. `editor._fl_styleWatch` mutates the third-party TinyMCE instance.** Prefer a `WeakMap<editor, watchRefs>` or closure-scoped object.

**8. `fl-typing` empty-check regex is fragile** (`clearTypingIfEmpty`) — `/<p>(?:\s|&nbsp;|<br\s*\/?>)*<\/p>/gi` won't match `<p style="...">` or `<p class="...">` which TinyMCE emits. Use `editor.dom.isEmpty(editor.getBody())` or a plain `textOnly.length === 0` check.

**9. Polling loop runs per-editor.** 6 seconds at 250ms × N rich text fields is a lot of sustained DOM work. Consider 3–4 attempts at increasing intervals (100ms / 500ms / 1500ms) to cover the common late-CSS case without the sustained pressure.

---

### Suggestions (Nice to Have)

- **Stop observing once styles are stable**: if the signature hasn't changed across N consecutive polls, disconnect the observers. Appearance rarely changes mid-session.
- **Consider CSS custom properties instead**: have the Appearance panel export computed values as CSS variables (`--fl-input-color`, `--fl-placeholder-color`, …) on the form root, and let the iframe consume them via `var(...)`. Removes the probe-and-sync dance entirely. Worth a follow-up ticket.
- **Branch target** (`projects/PS-1633`) — widget PRs typically merge to `master`. Confirm this is a long-lived release branch, otherwise re-target.
- **No tests added.** At least a Jest unit test for `collectEditorStyles()` (stubbed `getComputedStyle`) and `clearTypingIfEmpty()` (various raw HTML inputs) would give future edits a safety net.

---

### Verdict

**REQUEST_CHANGES**

The feature works, but the implementation is considerably heavier than the problem warrants. The three critical items — broad-scope `MutationObserver`, duplicated focus handler, untracked timers after teardown — are real correctness/performance risks that will show up in production as sluggish forms or Sentry noise. The warnings are cleanup items that are cheap to address while the code is fresh. Worth opening a follow-up ticket for the CSS-variables approach so this PR can be retired rather than maintained long-term.

zeryabkhan91 (approved):
LGTM

